PR #29, opened 2020-05-28
Fix incorrect exponent of number of MPR pairs in HistogramDisplay

### LINKED ISSUE #19: HistogramDisplay incorrect number of MPR pairs exponent
The y-axis of the histogram is not correct.  (It worked fine under Python 2, but seems to have broken in the move to Python 3.). The number of MPRs is 10^x where x is an exponent that is now always showing up as 0.  The problem appears to be in the way that exponent_text is extracted at line 40.

newsatit:
Problem: 
There is a bug in `plt.ticklabel_format` that does not update the yaxis offset text. Because we use the yaxis offset text to calculate the exponent, it cause the exponent to be calculated incorrectly.

Fix: 
Add `plt.tight_layout()` as a hack to make matplotlib recalculate and rerender the exponent when `ticklabel_format` runs as suggested here https://stackoverflow.com/a/31522781.

Resolves #19

83099220-5d48f600-a072-11ea-8663-f7f585ed2516

### DIFF
--- a/clumpr/HistogramDisplay.py
+++ b/clumpr/HistogramDisplay.py
@@ -36,7 +36,9 @@ def plot_histogram(plot_file, histogram, width, tree_name, d, t, l, max_x=None, 
     ax = plt.gca()
     # matplotlib sure is intuitive and easy to use!
     # Get the exponent text from the y-axis and format it into latex
-    plt.tight_layout() # A hack to make offset text updated (matplotlib bug) for this issue: https://github.com/ssantichaivekin/eMPRess/pull/29
+    # tight_layout is a hack that makes the exponent display correctly.
+    # Please refer to https://github.com/ssantichaivekin/eMPRess/pull/29
+    plt.tight_layout() 
     exponent_text = ax.get_yaxis().get_offset_text().get_text()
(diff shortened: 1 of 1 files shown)


clumpr/HistogramDisplay.py
```diff
@@ -36,6 +36,7 @@ def plot_histogram(plot_file, histogram, width, tree_name, d, t, l, max_x=None,
     ax = plt.gca()
     # matplotlib sure is intuitive and easy to use!
     # Get the exponent text from the y-axis and format it into latex
+    plt.tight_layout() # A hack to make offset text updated (matplotlib bug) for this issue: https://github.com/ssantichaivekin/eMPRess/pull/29
```
ssantichaivekin, line 39: Please move the inline comment to a separate line:
```
# tight_layout is a hack that makes the exponent display correctly.
# Please refer to https://github.com/ssantichaivekin/eMPRess/pull/29
plt.tight_layout()
```

newsatit:
Result
Screen Shot 2020-05-27 at 11 32 11 PM

ssantichaivekin:
Great job! Thanks for your contribution!

PR #33, opened 2020-05-31
First draft of empress functionality wrapper interface

ssantichaivekin:


empress_wrapper_examples.py
```diff
@@ -0,0 +1,74 @@
...
+import matplotlib.pyplot as plt
+import tkinter
+
+# Read Reconciliation Input
+recon_input = empress.read_tree("./example/heliconius.newick")
+
+# Display Costscape Window where we can click on x/y axis
+class TkinterCostPolygonPage:
+    """Put this class in your app"""
+    def __init__(self):
```
ssantichaivekin, line 14: I haven't tested whether something like this will work with retrieving the the x/y position on click.

empress_wrapper_examples.py
```diff
@@ -0,0 +1,74 @@
+# TODO: To be converted to automated tests
```
ssantichaivekin, line 1: This is only for showing what it will probably look like. Haven't implemented or tested the implementation.

empress/__init__.py
```diff
@@ -0,0 +1,102 @@
...
+        figure.savefig(fname)
+
+class ReconInputWrapper:
+    def __init__(self, parasite_tree: dict, host_tree: dict, tip_mapping: dict):
+        self._parasite_tree = parasite_tree
+        self._host_tree = host_tree
+        self._tip_mapping = tip_mapping
+
+class ReconciliationWrapper(Drawable):
+    # TODO: Replace dict with Reconciliation type
```
ssantichaivekin, line 44: Note that changing the internals of `ReconciliationWrapper` should not affect front end code.

empress_wrapper_examples.py
```diff
@@ -0,0 +1,74 @@
...
+        # or put these lines in your class
+        cost_poly = empress.compute_cost_polygon(recon_input)
+        fig = cost_poly.draw()
+        canvas = FigureCanvasTkAgg(fig)
+        canvas.draw()
+
+# Compute ReconGraph
+recongraph = empress.reconcile(recon_input, 1, 3, 2)
+
+# Save Histogram
```
ssantichaivekin, line 24: By save, I mean like draw it to file.

empress_wrapper_examples.py
```diff
@@ -0,0 +1,74 @@
...
+        canvas.draw()
+
+# Save Reconciliation drawing
+recon.draw_to_file("./recon_example.pdf")
+
+# Display Histogram of different clusters (by creating multiple matplotlib axes)
+class TkinterMultiHistogramPage:
+    """Put this class in your app"""
+    def __init__(self):
+        fig, axs = plt.subplots(3, 3)
```
ssantichaivekin, line 55: I think it's reasonable to ask front end developers to have a good understanding of matplotlib `Figure` and `Axes` and create the figures/axes instances themselves as shown in this line.

empress_wrapper_examples.py
```diff
@@ -0,0 +1,74 @@
...
+import empress
+from matplotlib.backends.backend_tkagg import FigureCanvasTkAgg
+import matplotlib.pyplot as plt
+import tkinter
+
+# Read Reconciliation Input
+recon_input = empress.read_tree("./example/heliconius.newick")
+
+# Display Costscape Window where we can click on x/y axis
+class TkinterCostPolygonPage:
```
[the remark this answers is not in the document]
rlhadas: I'd rename this TkinterCostRegionPage (since we use "regions" to describe these polygons in the xscape paper.

empress_wrapper_examples.py
rlhadas: I agree.  Note though that the subplots arguments won't necessarily be 3, 3.

empress/__init__.py
```diff
@@ -0,0 +1,102 @@
...
+    def find_median(self) -> ReconciliationWrapper:
+        """
+        Find and return one median of self
+        """
+        pass
+
+    def draw_on(self, axes: plt.Axes):
+        pass
+
+    def cluster(self, n) -> List['ReconGraphWrapper']:
```
rlhadas, line 67: Our clustering algorithm requires some input arguments (e.g., depth or nMPRs).

empress/__init__.py
```diff
@@ -0,0 +1,102 @@
...
+        """
+        Cluster self into n reconciliation graphs
+        """
+        pass
+
+class CostPolygonWrapper(Drawable):
```
rlhadas, line 73: Not sure what this one does, but I'd call this CostRegionWrapper rather than CostPolygonWrapper

empress/__init__.py
```diff
@@ -0,0 +1,102 @@
...
+        pass
+
+def read_tree(fname: str) -> ReconInputWrapper:
+    """
+    Read parasite tree, host tree, and tip mapping from fname.
+    Returns ReconInputWrapper which wraps the three info.
+    """
+    pass
+
+def compute_cost_polygon(recon_input: ReconInputWrapper) -> CostPolygonWrapper:
```
rlhadas, line 90: Replace "polygon" with "region"

empress_wrapper_examples.py
ssantichaivekin: I agree. `(3,3)` is only an example :)

empress/__init__.py
ssantichaivekin: I think it the reconciliation graph should keep track of arguments that can be confusing to users. I think the only required argument should be the number of clusters the user wants (shown in the argument as `n`).

empress/__init__.py
ssantichaivekin: ok. It's a wrapper that draws the costscape window :)

empress/__init__.py
ssantichaivekin: I will also add other arguments (because I need to haha) when I actually implement the function :).

empress_wrapper_examples.py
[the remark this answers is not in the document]
ssantichaivekin: renamed! :) (see new file changes!)

empress/__init__.py
ssantichaivekin: renamed! :) (see new file changes!)

empress/__init__.py
ssantichaivekin: renamed! :) (see new file changes!)

empress/__init__.py
rlhadas: I agree that the only argument here should be the number of clusters.  The hidden parameters can go someplace else.  I don't think it should be in the reconciliation graph, but we can decide on that later when Ross has had a chance to work on how these hidden parameter values get chosen.

empress/__init__.py
```diff
@@ -0,0 +1,99 @@
...
+    def __init__(self, info):
+        """
+        Will look into this more.
+        """
+        pass
+
+    def draw_on(self, axes: plt.Axes):
+        pass
+
+def read_tree(fname: str) -> ReconInputWrapper:
```
rlhadas, line 80: This might better be name read_input since it's not just a single tree that is being read here.

empress/__init__.py
```diff
@@ -0,0 +1,99 @@
...
+    """
+    Read parasite tree, host tree, and tip mapping from fname.
+    Returns ReconInputWrapper which wraps the three info.
+    """
+    pass
+
+def compute_cost_region(recon_input: ReconInputWrapper) -> CostRegionWrapper:
```
rlhadas, line 87: I don't know what the use case is for this one, but that's OK.  We can leave it here for now.

ssantichaivekin: This runs the reconcile algorithm in costcape behind the scene. It produces a list of cost vectors (and other information) which can be used for visualization.

empress/__init__.py
ssantichaivekin: Will do.

empress/__init__.py
ssantichaivekin: This is running costscape behind the scene. It produces a list of cost vectors and wraps it in the cost region wrapper.

PR #35, opened 2020-06-02
New empress.py file and entire xscape directory.

rlhadas:
xscape code modified so costscape fixes loss cost at 1.  Duplication and transfer costs are now relative to loss cost.  This resulted in changes to a number of files in the xscape directory.  Also updated empress.py to reflect that change.  empress.py was also restricting dtl costs to be integers and this was changed to float.  Ultimately, we may want to change that to our own Rational number class to simulate floats but avoiding roundoff error.


plotcostsAnalytic.py
```diff
@@ -0,0 +1,118 @@
+# plotcostsAnalytic.py
```
ssantichaivekin, line 1: I noticed that this file just got added to the folder in the root repository. Is it supposed to be here?

rlhadas: Sorry.  It should replace the function with the same name in xscape.  I've updated that from the original pull request to fix a bug in the way that the legend was displayed.

xscape/plotcostsAnalytic.py
```diff
@@ -1,6 +1,7 @@
 # plotcostsAnalytic.py
```
ssantichaivekin, line 1: It seems that you are also modifying the `plotcostsAnalysic.py` file here. I think you want to modify this file but not add a new file?

plotcostsAnalytic.py
ssantichaivekin: I think you have already modified the file at `xscape/plotcostsAnalytic.py`.

ssantichaivekin: Is that the same file as this one?

empress.py
```diff
@@ -41,52 +42,52 @@ def process_arg():
...
     score.add_argument("--pdv", action="store_true",
-        help="Use the weighted average distance to evaluate clusters.")
+        help="Use the weighted average distance to evaluate clusters")
     score.add_argument("--support", action="store_true",
-        help="Use the weighted average event support to evaluate clusters.")
+        help="Use the weighted average event support to evaluate clusters")
 
     ### Parser for Histogram ###
     histogram_parser = subparsers.add_parser('histogram', help="Run histogram")
-    histogram_parser.add_argument("-d", type=int, metavar="<duplication_cost>",
```
ssantichaivekin, line 84: I think these changes should be in a different pull request.

empress.py
```diff
@@ -41,52 +42,52 @@ def process_arg():
...
     depth_or_n.add_argument("--depth", type=int, metavar="<tree_depth>",
-        help="How far down the graph to consider event splits.")
+        help="How far down the graph to consider event splits")
     depth_or_n.add_argument("--nmprs", type=int, metavar="<tree_depth>",
-        help="How many MPRs to consider")
+        help="Number of MPRs to consider")
     # What visualizations to produce
     vis_type = clumpr_parser.add_mutually_exclusive_group(required=False)
     vis_type.add_argument("--pdv-vis", action="store_true",
-        help="Visualize the resulting clusters using the PDV.")
```
ssantichaivekin, line 72: I think these changes should be in a different pull request.

PR #44, opened 2020-06-04
Add Monte Carlo empirical p-value computation and visualization for Reconciliation

rlhadas:
stats function computes empirical p-value
plot_histogram plots the statistics histogram


ssantichaivekin:
Hi Prof Ran, could you rename this to something like "Add p-value testing and visualization for Reconciliation"?

empress/clumpr/Statistics.py
```diff
@@ -0,0 +1,98 @@
...
+    :param phi: dictionary representation of parasite tip to host tip mapping
+    :param dup_cost: float duplication cost
+    :param transfer_cost: float transfer cost
+    :param loss_cost: float loss cost
+    :param num_trials: int number of trials in Monte Carlo simulation
+    :return: list of floating point costs of reconciliations of the Monte Carlo samples
+    """
+   
+    costs = list()  # List of costs of random trials
+    parasites = list(phi.keys())
```
ssantichaivekin, line 30: You don't need to convert it to list if you are not going to access it by index. `phi.keys()` are already iterable.

empress/clumpr/Statistics.py
```diff
@@ -0,0 +1,98 @@
...
+    hosts = list(phi.values())
```
ssantichaivekin, line 31: same here

empress/clumpr/Statistics.py
```diff
@@ -0,0 +1,98 @@
...
+        if h in parasite_count: parasite_count[h] += 1
+        else: parasite_count[h] = 1
+    for h in unique_hosts:
+        parasite_samples = random.sample(parasites, parasite_count[h])
+        for p in parasite_samples: 
+            random_phi[p] = h
+            parasites.remove(p)
+    return random_phi
+
+def plot_histogram(MPR_cost: int, costs: list):
```
ssantichaivekin, line 64: I think the variable name should be snake_case. Could you change this to `mpr_cost`?

ssantichaivekin:
Could you explain a little bit more about what you mean by statistics in "stats function computes statistics"? Feel free to add screenshots from Jane to this pull request if you find it helpful.

empress/clumpr/Statistics.py
```diff
@@ -0,0 +1,98 @@
...
+    :param MPR_cost: floating point cost of MPR for given host-parasite-phi and DTL costs
+    :param costs: list of floating point costs of Monte Carlo samples
+    :return: None.  Displays histogram.
+    """
+    plt.hist(costs, color = "blue", density = True)
+    plt.hist(MPR_cost, color="red", rwidth=1)
+    plt.show()
+
+def stats(host_tree: dict, parasite_tree: dict, phi: dict, dup_cost: float, transfer_cost: float, loss_cost: float, num_trials: int) -> (float, list, float):
+    _, MPR_cost, _, _ = DTLReconGraph.DP(host_tree, parasite_tree, phi, dup_cost, transfer_cost, loss_cost)
```
ssantichaivekin, line 75: This line of code should not go before the comment.

ssantichaivekin: I feel like this function should not call `DTLReconGraph.DP` to get the `MPR_cost` (which I think should be renamed to `mpr_cost`). I think the users of this function should call the DP algorithm directly. If we put the DP algorithm here, the DP algorithm might get called multiple times (once for finding the reconciliation tree, and another time for finding the statistics).

empress/clumpr/Statistics.py
rlhadas: Fixed.

empress/clumpr/Statistics.py
rlhadas: This once is necessary because we are using random.choice(hosts).

empress/clumpr/Statistics.py
rlhadas: Fixed.

empress/clumpr/Statistics.py
rlhadas: Moved line before the comment and changed MPR_cost to mpr_cost.  I'm not sure how to resolve the DP issue.  Each trial needs to re-run the DP anew.  It's true that the DP gives us more than what we need here, but we do need to recompute the DP cost each time.  Please feel free to offer another solution - I just don't see it yet. ;^)

Statistics.py
```diff
@@ -0,0 +1,102 @@
...
+    :param loss_cost: float loss cost
+    :param num_trials: int number of trials in Monte Carlo simulation
+    :param mpr_cost: float cost of the mpr from the original dataset.  If None, we compute it here
+    :return: tuple of three items:
+        float cost of optimal MPR for given data
+        list of floating point costs of reconciliations of the Monte Carlo samples
+        float empirical p-value between 0 and 1
+    """
+    # If mpr_cost not passed in, we compute it here (only used for ease of testing)
+    if mpr_cost is None:
```
ssantichaivekin, line 91: I think it's okay to remove `mpr_cost` param and just compute the DP every time like you did previously. I feel like that is cleaner than this. (sorry for a bad suggestion.) Are you planning to just export the stats function and keep the others internal?

Statistics.py
```diff
@@ -0,0 +1,102 @@
...
+# Ran Libeskind-Hadas, June 2020
+
+# Performs Monte Carlo simulation to compute empirical p-value and histogram
+# of solution costs.
+# https://www.ncbi.nlm.nih.gov/pmc/articles/PMC379178/
+
+from clumpr import DTLReconGraph
+import random
+import matplotlib
+matplotlib.use('tkagg') # need this so plt.show() works
```
ssantichaivekin, line 12: Please remove this `matplotlib.use('tkagg')` line as it will be addressed in #49, #52 .

Statistics.py
```diff
@@ -0,0 +1,102 @@
...
+import matplotlib.pyplot as plt
+
+from typing import List, Dict, Tuple
```
ssantichaivekin, line 15: The `List`, `Dict`, and `Tuple` type from typing is used for defining nested types (say, `Dict[str, int]`, `Tuple[str, int]`). It seems that you are not using these in your code, so feel free to remove this line. Note that if you are only using `dict`, `float` and other basic python types for type checking, you don't need to import `typing`.

Statistics.py
```diff
@@ -0,0 +1,102 @@
...
+
+def trials(host_tree: dict, parasite_tree: dict, phi: dict, dup_cost: float, transfer_cost: float, loss_cost: float, num_trials: int) -> list:
```
ssantichaivekin, line 17: Do you want other files to call this function `trails` and `create_random_phi`? Or do you want it to be internal? You can tell the users of this file that some functions are internal by naming functions with a leading underscore `_trails` and `_create_random_phi` or you can create an `__all__` variable at the top of the module which lists all functions you intend to export. The functions with leading underscore or not in all variable will not be suggested to users by autocomplete tools such as vscode python extension and will not be documented by document generation tools such as sphinx.

PR #54, opened 2020-06-05
Refactor import to conform with Google Python Guideline

### LINKED ISSUE #6: Refactor empress to use absolute import path
Currently, we are importing files(modules) in a non-standard way. We want to either use [absolute import or relative import](https://realpython.com/absolute-vs-relative-python-imports/). According to Google Python style guide, [absolute import is preferred](https://realpython.com/absolute-vs-relative-python-imports/). This change involves modifying **all** import lines in the codebase.

Note, we have decided to use absolute import.

Examples:

Before (in `clumpr/DTLReconGraph.py`):
```python
import Greedy
import newickFormatReader
import ReconcileMainInput
```
After absolute import:
```python
from clumpr import Greedy
from clumpr import newickFormatReader
from clumpr import ReconcileMainInput
```

Before (in `xcape/costcape.py`):
```python
try:
    import xscape
except ImportError:
    import sys
    from os.path import realpath, dirname, join
    sys.path.append("../python")
    import xscape
from xscape.commonAnalytic import *
from xscape.CostVector import *
from xscape import getInput
```
After absolute import:
```python
from xscape.python.xscape.commonAnalytic import *
from xscape.python.xscape.costVector import *
from xscape.python.xscape import getInput
```

Pros of using either absolute import or relative import:
- Allow editors such as vscode and pycharm to see and analyze imports

Pros of using absolute import:
- Easier to write unit tests
- Easier to read than relative import
- Easier to refactor if we change the locations of folders
- Consistent with Google style guide

Cons of using absolute import:
- We will not be able to run files in subdirectories independently. `python clumpr/DTLReconGraph.py trees.newick 1 2 1` will no longer work (see comment). We will have to write `python -m clumpr.DTLReconGraph trees.newick 1 2 1` instead for the same functionality. This is less intuitive for newcomers. However, I think we should remove the option to run each file on its own anyway, so I think this is warranted

Cons of not switching:
- Hard to navigate code in editors because `ctrl` + `click` does not work
- Hard to write unit tests in a different folder far from the folder being tested
- Hard to use tools that analyze the relationships between different files in the code

maxoja:
Resolves #6 .


ssantichaivekin:
Have you run the commands on the readme and tested that they still work?

maxoja:
I tested those that works at when I got onboard along with main of submodules and some test scirpts that are runnable. I think we should wait for the fix first before merge

ssantichaivekin:
I think the issue is already fixed. Feel free to merge from master.

empress/xscape/eventscape.py
```diff
@@ -5,21 +5,16 @@
 
 # python libraries
 import time
+import csv
 from collections import *
```
maxoja, line 9: I found that removing this import passes all the provided script runs. However, I'm not 100% sure that this isn't use in the script. Do you want me to remove it and see if it will break at some point later? I think the error should be obvious when arose if existed. @ssantichaivekin 

This comment is as part of PR #54

maxoja: Sry, I pin the wrong line. I meant to pin line 9 ...

ssantichaivekin: You can also search the file whether the name `csv` is mentioned somewhere. If not remove it. Some IDEs such as PyCharm will tell you immediately when an import is unused.

maxoja: Jo, sry but I meant to pin `from collections import *` and I use VS Code without proper extension to track if any of the code in the file use anything from collections module.

ssantichaivekin: Oh my bad.

ssantichaivekin: I also think we should remove import *, but we can do that later :)

ssantichaivekin:
Have you tested the functions after merging from master?

maxoja:
Yes I have tested all the script runs in readme.md

PR #58, opened 2020-06-05
Modify visualization for ReconGraph to save as .gv if Graphviz not installed

ssantichaivekin:
Modify ReconGraph visualization to save as .gv (DOT file for Graphviz) in case Graphviz is not installed in path and the user cannot render the image as .png. 

Note that the file doesn't change much. It's just that the file used a different new line character (CRLF) so the diff looks a little weird. This is probably because the file was created in Windows.


rlhadas:
My sense is that while this is a useful tool for doing theory/algorithms development, it is not likely to be used by the empress user community and I suggest that it either not be shipped with empress or it be placed in a special package like research_tools or misc.

rlhadas:
Looks good.  Thansk!

ssantichaivekin:
I think it's cleaner to have this as function be a method of the class `ReconGraph`. The cli and gui may or may not call it, but I think it makes sense for the function to be here in `ReconGraph` and in the documentation of `ReconGraph`.

PR #60, opened 2020-06-06
Greedy.py is no longer used.

rlhadas:


ssantichaivekin:
It turns out that `Greedy.py` is still in used. We have this error because we don't have unit tests in place to detect this.

PR #63, opened 2020-06-08
Add GUI

### LINKED ISSUE #38: Merge UI code to this repository
Merge tkinter GUI code from [Qing-Yang-1211/eMPRess-GUI](https://github.com/Qing-Yang-1211/eMPRess-GUI) to this repository.

Qing-Yang-1211:
Resolves #38 .
matplotlib click
string invalid input
negative invalid input
valid input


eMPRess_GUI_version1.py
```diff
@@ -0,0 +1,354 @@
+# eMPRess_GUI_iteration5
```
ssantichaivekin, line 1: Please rename this file to `empress_gui.py`

eMPRess_GUI_version1.py
```diff
@@ -0,0 +1,354 @@
...
+import matplotlib.pyplot as plt
+from matplotlib.backends.backend_tkagg import FigureCanvasTkAgg, NavigationToolbar2Tk
+from matplotlib.backend_bases import MouseEvent, key_press_handler
+import empress
+
+class App:
+
+    def __init__(self, master):
+        self.master = master
+        ### configures the master frame ###
```
ssantichaivekin, line 16: Please use one `#` for block comments: `# configures the master frame`.

eMPRess_GUI_version1.py
```diff
@@ -0,0 +1,354 @@
...
+        self.logo_frame.grid_propagate(False)
+
+        # adds background image
+        photo = tk.PhotoImage(file="JaneTransLogo-thin.png")
+        label = tk.Label(self.logo_frame, image=photo)
+        label.place(x=0, y=0)
+        label.image = photo
+
+        ### creates an input frame on the left side of the master frame ###
+        self.func_frame = tk.Frame(master)
```
ssantichaivekin, line 34: Why is this variable called `func_name`?

eMPRess_GUI_version1.py
```diff
@@ -0,0 +1,354 @@
...
+        self.func_frame.grid_rowconfigure(0, weight=1)
+        self.func_frame.grid_rowconfigure(1, weight=1)
+        self.func_frame.grid_rowconfigure(2, weight=1)
+        self.func_frame.grid_rowconfigure(3, weight=1)
+        self.func_frame.grid_columnconfigure(0, weight=1)
+        self.func_frame.grid_propagate(False)
+
+        ### creates an output frame on the right side of the master frame ###
+        self.output_frame = tk.Frame(master)
+        self.output_frame.grid(row=1, column=1, sticky="nsew")
```
ssantichaivekin, line 45: What does sticky mean? If the meaning is not obvious, it might be a good idea to add a comment.

eMPRess_GUI_version1.py
```diff
@@ -0,0 +1,354 @@
...
+
+        ### "Compute Reconciliations" button ###
+        # displays reconciliation results(numbers) and three options(checkboxes) for viewing graphical analysis
+        self.compute_recon_button = tk.Button(self.func_frame, text="Compute Reconciliations", command=self.recon_analysis, state=tk.DISABLED)
+        self.compute_recon_button.grid(row=2, column=0)
+
+    def load_file(self):
+        """Loads in an input file and displays the number of leaves in each tree when "Load File" button is clicked."""
+        # creates a frame to display the number of leaves and the file path
+        recon_input_info_frame = tk.Frame(self.output_frame)
```
ssantichaivekin, line 73: I think you can rename `recon_input_info_frame` to `input_info_frame` because our code only has one kind of input. I think in vscode you can do right click -> refactor -> rename, or you can also use find and replace, but don't trust the functionalities too much.

eMPRess_GUI_version1.py
```diff
@@ -0,0 +1,354 @@
...
+        recon_input_info_frame.grid(row=0, column=0, sticky="nsew")
+        recon_input_info_frame.grid_columnconfigure(0, weight=1)
+        recon_input_info_frame.grid_rowconfigure(0, weight=1)
+        recon_input_info_frame.grid_rowconfigure(1, weight=1)
+        recon_input_info_frame.grid_rowconfigure(2, weight=1)
+        recon_input_info_frame.grid_propagate(False)
+
+        # allows loading a .newick file
+        self.file_path = ""
+        input_file = tk.filedialog.askopenfilename(initialdir="/desktop", title="Select a file")
```
ssantichaivekin, line 83: Is there a way to select `initialdir` to the current directory that the user is running the script from? I think that might be a better default than "desktop". If you could think of a better default, feel free to use it. Desktop is okay too I think. On a side note, does using `/desktop` as the initial directory works? For example, my desktop is located at `~/desktop` which is short for `/Users/santisantichaivekin/desktop` but not at `/desktop`.

eMPRess_GUI_version1.py
```diff
@@ -0,0 +1,354 @@
...
+        recon_input_info_frame = tk.Frame(self.output_frame)
+        recon_input_info_frame.grid(row=0, column=0, sticky="nsew")
+        recon_input_info_frame.grid_columnconfigure(0, weight=1)
+        recon_input_info_frame.grid_rowconfigure(0, weight=1)
+        recon_input_info_frame.grid_rowconfigure(1, weight=1)
+        recon_input_info_frame.grid_rowconfigure(2, weight=1)
+        recon_input_info_frame.grid_propagate(False)
+
+        # allows loading a .newick file
+        self.file_path = ""
```
ssantichaivekin, line 82:
```python
self.file_path = None
```

eMPRess_GUI_version1.py
```diff
@@ -0,0 +1,354 @@
...
+            # this is only DEMO for now
+            host_tree_info = tk.Label(recon_input_info_frame, text="Host:  83 tips (DEMO)")
+            host_tree_info.grid(row=1, column=0, sticky="w")
+            parasite_info = tk.Label(recon_input_info_frame, text="Parasite/symbiont:  78 tips (DEMO)")
+            parasite_info.grid(row=2, column=0, sticky="w")           
+        else:
+            messagebox.showinfo("Warning", "Please load a '.newick' file.")
+
+        # enables the next step, setting DTL costs
+        if self.file_path != "":
```
ssantichaivekin, line 98:
```python
if self.file_path is not None:
```

eMPRess_GUI_version1.py
ssantichaivekin: It's good practice to use `None` as default value.

eMPRess_GUI_version1.py
```diff
@@ -0,0 +1,354 @@
...
+        trans_entry.bind('<Return>', self.check_trans_input)
+        loss_entry.bind('<Return>', self.check_loss_input)
+        #dup_entry.bind('<FocusOut>', self.check_dup_input)  # (bug) this will make the invalid input warning appear twice
+        #trans_entry.bind('<FocusOut>', self.check_trans_input)  # same as above 
+        #loss_entry.bind('<FocusOut>', self.check_loss_input)  # same as above 
+        dup_entry.grid(row=0, column=1, sticky="w")
+        trans_entry.grid(row=1, column=1, sticky="w")
+        loss_entry.grid(row=2, column=1, sticky="w")
+
+    def plot_cost_landscape(self):
```
ssantichaivekin, line 136: I think this is called `cost_regions` in other parts of the code. Use this name convention over `cost_landscape`

ssantichaivekin: Also modify the comments.

eMPRess_GUI_version1.py
```diff
@@ -0,0 +1,354 @@
...
+        plt_frame.pack_propagate(False)
+        # use empress wrapper
+        recon_input = empress.read_input(self.file_path)
+        cost_regions = empress.compute_cost_regions(recon_input, 0.5, 10, 0.5, 10)  
+        #cost_regions.draw_to_file('./examples/cost_poly.png')  # draw and save to a file
+        fig = cost_regions.draw()  # draw to figure (creates matplotlib figure)
+        canvas = FigureCanvasTkAgg(fig, plt_frame)
+        canvas.draw()
+        canvas.get_tk_widget().pack(side=tk.TOP, fill=tk.BOTH, expand=True)
+        toolbar = NavigationToolbar2Tk(canvas, plt_frame)
```
ssantichaivekin, line 154: Does this toolbar allow users to zoom in/out? It might be a good idea to add a small comment what the toolbar enables the user to do.

eMPRess_GUI_version1.py
```diff
@@ -0,0 +1,354 @@
...
+        """Plots the cost landscape using matplotlib and embeds the graph in a tkinter window."""    
+        # creates a new tkinter window 
+        plt_window = tk.Toplevel(self.master)
+        plt_window.geometry("550x550")
+        plt_window.title("Matplotlib Graph DEMO")
+        # creates a new frame
+        plt_frame = tk.Frame(plt_window)
+        plt_frame.pack(fill=tk.BOTH, expand=1)
+        plt_frame.pack_propagate(False)
+        # use empress wrapper
```
ssantichaivekin, line 146: Remove this line. If a person wonders what it does, they can hover over the code to see description or command + click on the code.

eMPRess_GUI_version1.py
```diff
@@ -0,0 +1,354 @@
...
+    def check_dup_input(self, *args):
+        """Checks and updates the duplication cost when user enters a value in the entry box."""
+        dupValue_without_decimal_point = self.dup_cost.get().replace(".", "")
+        if str.isdigit(dupValue_without_decimal_point) and ("-" not in dupValue_without_decimal_point):
+                self.dup_cost.set(str(round(float(self.dup_cost.get()), 2)))
+                self.has_dup_cost = True
+                self.enable_recon_btn()
+        else:
+            messagebox.showinfo("Warning", "Please enter a non-negative number.")
+            self.dup_cost.set("")
```
ssantichaivekin, line 189: Is there a way to use `None` instead of `""` as default arguments in case things does not exist?

eMPRess_GUI_version1.py
```diff
@@ -0,0 +1,354 @@
...
+            messagebox.showinfo("Warning", "Please click inside the axes bounds.")
+
+    def enable_recon_btn(self):
+        """Enables the next step, viewing reconciliation results, after DTL costs are set correctly."""
+        if self.has_dup_cost == True and self.has_trans_cost == True and self.has_loss_cost == True:
+            self.compute_recon_button.configure(state=tk.NORMAL)
+
+    def check_dup_input(self, *args):
+        """Checks and updates the duplication cost when user enters a value in the entry box."""
+        dupValue_without_decimal_point = self.dup_cost.get().replace(".", "")
```
ssantichaivekin, line 182: I'm not sure what you are doing here. If you want to round the decimal off, convert it to `int`, if you want to round to the nearest 2nd decimal place, see https://docs.python.org/3/library/functions.html#round and https://www.tutorialspoint.com/python3/number_round.htm . The code should look something like this:
```python
x = "1.732"
y = float(x)
z = round(y, 2) # now y = 1.73
if z <= 0:
    print("the number is negative!")
```

eMPRess_GUI_version1.py
```diff
@@ -0,0 +1,354 @@
...
+        if str.isdigit(dupValue_without_decimal_point) and ("-" not in dupValue_without_decimal_point):
+                self.dup_cost.set(str(round(float(self.dup_cost.get()), 2)))
```
ssantichaivekin, line 184: Is there a way to save the variable as a float and not a string?

eMPRess_GUI_version1.py
```diff
@@ -0,0 +1,354 @@
...
+                self.has_dup_cost = True
```
ssantichaivekin, line 185: I think we could check whether we have duplication cost by checking whether
```python
self.dup_cost is not None
```

eMPRess_GUI_version1.py
```diff
@@ -0,0 +1,354 @@
...
+                self.enable_recon_btn()
```
ssantichaivekin, line 186: Do we have to check that the three values are there before we enable the button?

eMPRess_GUI_version1.py
```diff
@@ -0,0 +1,354 @@
...
+        else:
+            messagebox.showinfo("Warning", "Please enter a non-negative number.")
+            self.dup_cost.set("")
+            self.has_dup_cost = False
+
+    def check_trans_input(self, *args):
```
ssantichaivekin, line 192: Same here.

eMPRess_GUI_version1.py
```diff
@@ -0,0 +1,354 @@
...
+
+
+root = tk.Tk()
+root.geometry("600x600")
+root.title("eMPRess GUI Version 1")
+App(root)
+root.protocol("WM_DELETE_WINDOW", on_closing)
+root.mainloop()
+root.quit()
+
```
ssantichaivekin, line 325: remove extra spaces

eMPRess_GUI_version1.py
```diff
@@ -0,0 +1,354 @@
...
+        if str.isdigit(transValue_without_decimal_point) and ("-" not in transValue_without_decimal_point):
+                self.trans_cost.set(str(round(float(self.trans_cost.get()), 2)))
+                self.has_trans_cost = True
+                self.enable_recon_btn()
+        else:
+            messagebox.showinfo("Warning", "Please enter a non-negative number.")
+            self.trans_cost.set("")
+            self.has_trans_cost = False
+    
+    def check_loss_input(self, *args):
```
ssantichaivekin, line 204: Same here.

eMPRess_GUI_version1.py
```diff
@@ -0,0 +1,354 @@
...
+        recon_space_btn = tk.Checkbutton(recon_checkbox_frame, text="View solution space", padx=10, variable=self.recon_space_btn_var, command=self.open_and_close_window_recon_space)
+        recons_btn = tk.Checkbutton(recon_checkbox_frame, text="View reconciliations", padx=10, variable=self.recons_btn_var, command=self.open_and_close_window_recons)
+        histogram_btn = tk.Checkbutton(recon_checkbox_frame, text="Stats mode", padx=36, variable=self.histogram_btn_var, command=self.open_and_close_window_histogram)
+        recon_space_btn.pack()
+        recons_btn.pack()
+        histogram_btn.pack()
+
+        # shows reconciliation results as numbers
+        recon_nums_frame = tk.Frame(self.output_frame)
+        recon_nums_frame.grid(row=2, column=0, sticky="nsew")
```
ssantichaivekin, line 234: Comment on one of the stickies to say what it does

eMPRess_GUI_version1.py
```diff
@@ -0,0 +1,354 @@
...
+### View p-value histogram ###
+class Histogram_window:
+    def __init__(self, master):
+        self.master = master        
+        self.frame = tk.Frame(master)
+        self.frame.pack(fill=tk.BOTH, expand=1)
+        self.frame.pack_propagate(False)
+
+def on_closing():
+   plt.close("all")
```
ssantichaivekin, line 314: It would be nice if you create an issue that explains why this line is necessary and close the issue, and then put the issue number or the link to the issue on this line. Alternatively, you could also write why you need to do this on this line.

eMPRess_GUI_version1.py
```diff
@@ -0,0 +1,354 @@
...
+        master.grid_columnconfigure(0, weight=1)
+        master.grid_columnconfigure(1, weight=1)
+
+        ### creates a logo frame on top of the master frame ###
+        self.logo_frame = tk.Frame(master)
+        self.logo_frame.grid(row=0, column=0, columnspan=2, sticky="nsew")
+        self.logo_frame.grid_propagate(False)
+
+        # adds background image
+        photo = tk.PhotoImage(file="JaneTransLogo-thin.png")
```
ssantichaivekin, line 28: move to `assets/file_name.png`.

eMPRess_GUI_version1.py
Qing-Yang-1211: Done:)

eMPRess_GUI_version1.py
Qing-Yang-1211: Done:)

eMPRess_GUI_version1.py
Qing-Yang-1211: It is named func_frame, i.e. function frame because it is the frame for those buttons (functionalities) on the left. There are also logo frame and the output frame. Or I can change this to input frame (input_frame). I think it will be better. :)

eMPRess_GUI_version1.py
Qing-Yang-1211: It means that self.output_frame expands in all four directions (north, south, east and west) to fully occupy the allocated space in the grid system (row 1 column 1). I will add the comment. :)

eMPRess_GUI_version1.py
Qing-Yang-1211: Done:)

Qing-Yang-1211: command+fn+f2 works:)

eMPRess_GUI_version1.py
Qing-Yang-1211: Done:)

eMPRess_GUI_version1.py
Qing-Yang-1211: Done:)

eMPRess_GUI_version1.py
Qing-Yang-1211: Done:)

eMPRess_GUI_version1.py
Qing-Yang-1211: Done:)

eMPRess_GUI_version1.py
Qing-Yang-1211: Yes, it does. I will add the comment. The toolbar allows the user to zoom in/out, drag the graph and save the graph.

eMPRess_GUI_version1.py
Qing-Yang-1211: I wanted to use str.isdigit() to check if the input contains only numbers instead of letters, but this function treats the decimal point as invalid input, so I removed the decimal point if it was there before using str.isdigit().

eMPRess_GUI_version1.py
Qing-Yang-1211: Yes. The compute reconciliations button should only be enabled (to display clumpers and stats mode and stuff) after the user has set the DTL costs correctly. :)

eMPRess_GUI_version1.py
Qing-Yang-1211: Done:) (for the logo_frame when sticky first appeared)

eMPRess_GUI_version1.py
Qing-Yang-1211: Done:)

eMPRess_GUI_version1.py
Qing-Yang-1211: I tried this. If I don't set any value for an empty entry box, its value will be "" instead of None. Then I went to initialize all entry boxes to be None, and I realized that each entry box will actually display "None" at the very beginning. What do you think? :))

Qing-Yang-1211: Oh, actually, this will still enable the next button if the "None" is not changed to a valid number. (I checked for self.xx_cost is not None for all three).

eMPRess_GUI_version1.py
Qing-Yang-1211: The comment was added for the logo_frame when sticky first appeared. :)

eMPRess_GUI_version1.py
Qing-Yang-1211: Please see the comments below for the same issue:)

eMPRess_GUI_version1.py
Qing-Yang-1211: Done:)

eMPRess_GUI_version1.py
Qing-Yang-1211: A comment was added for the on_closing function. It kills the matplotlib program and all other tkinter programs when the master window is closed.

ssantichaivekin: Could you create an issue for this or write a comment that explains what will happen if you remove this line? Something like `# If we don't run plt.close("all"), then python will not terminate after the gui exits`. Alternatively, write something like this as a comment in the file.

eMPRess_GUI_version1.py
Qing-Yang-1211: It didn't work I think. I used "import os" and "os.getcwd()" to set to the current working directory. I feel that we can also set it to desktop in this way.

eMPRess_GUI_version1.py
```diff
@@ -0,0 +1,354 @@
...
+        costs_frame.grid_rowconfigure(2, weight=1)
+        costs_frame.grid_propagate(False)
+
+        dup_label = tk.Label(costs_frame, text="Duplication:")
+        trans_label = tk.Label(costs_frame, text="Transfer:")
+        loss_label = tk.Label(costs_frame, text="Loss:")
+        dup_label.grid(row=0, column=0, sticky="w")
+        trans_label.grid(row=1, column=0, sticky="w")
+        loss_label.grid(row=2, column=0, sticky="w")
+        self.dup_cost = tk.StringVar()
```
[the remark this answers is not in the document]
Qing-Yang-1211: I have set it to a DoubleVar() variable. If the user input contains letters(cannot be converted to a float), a warning message window will pop up, saying "Please enter a non-negative number". If the user input is negative, the same warning message window will also pop up. Only when the three DTL costs have been set correctly, the next button will be enabled. All valid inputs will be displayed with 2 decimal places. DTL costs retrieved from matplotlib will be automatically displayed wit 2 decimal places.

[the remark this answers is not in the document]
Qing-Yang-1211: I have sent the code to you on slack if you would like to try it. :))

empress_gui.py
```diff
@@ -0,0 +1,335 @@
...
+    root.destroy()
+
+
+root = tk.Tk()
+root.geometry("600x600")
+root.title("eMPRess GUI Version 1")
+App(root)
+root.protocol("WM_DELETE_WINDOW", on_closing)
+root.mainloop()
+root.quit()
```
ssantichaivekin, line 335: Add one trailing new line. (see https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline)

empress_gui.py
```diff
@@ -0,0 +1,335 @@
...
+import matplotlib.pyplot as plt
+from matplotlib.backends.backend_tkagg import FigureCanvasTkAgg, NavigationToolbar2Tk
+from matplotlib.backend_bases import MouseEvent, key_press_handler
+import empress
+
+class App:
+
+    def __init__(self, master):
+        self.master = master
+        # configures the master frame
```
ssantichaivekin, line 17: I think it might look better to use capital letter to start a block comment.

empress_gui.py
```diff
@@ -0,0 +1,335 @@
...
+    def recon_analysis(self):
+        """Displays reconciliation results in numbers and further viewing options for graphical analysis."""
+        # creates a frame for the three checkbuttons
+        recon_checkbox_frame = tk.Frame(self.func_frame)
+        recon_checkbox_frame.grid(row=3, column=0, sticky="nsew")
+        recon_checkbox_frame.pack_propagate(False)
+        self.recon_space_btn_var = tk.IntVar()
+        self.recons_btn_var = tk.IntVar()
+        self.histogram_btn_var = tk.IntVar()
+        recon_space_btn = tk.Checkbutton(recon_checkbox_frame, text="View solution space", padx=10, variable=self.recon_space_btn_var, command=self.open_and_close_window_recon_space)
```
ssantichaivekin, line 235: Write this in multiple lines.
```python
recon_space_btn = tk.Checkbutton(recon_checkbox_frame, text="View solution space", 
    padx=10, variable=self.recon_space_btn_var,
    command=self.open_and_close_window_recon_space)
```
Something like this.

empress_gui.py
```diff
@@ -0,0 +1,335 @@
...
+        recons_btn = tk.Checkbutton(recon_checkbox_frame, text="View reconciliations", padx=10, variable=self.recons_btn_var, command=self.open_and_close_window_recons)
```
ssantichaivekin, line 236: Write this in multiple lines.

empress_gui.py
```diff
@@ -0,0 +1,335 @@
...
+        histogram_btn = tk.Checkbutton(recon_checkbox_frame, text="Stats mode", padx=36, variable=self.histogram_btn_var, command=self.open_and_close_window_histogram)
```
ssantichaivekin, line 237: Write this in multiple lines.

empress_gui.py
```diff
@@ -0,0 +1,335 @@
...
+
+    def __init__(self, master):
+        self.master = master
+        # configures the master frame 
+        master.grid_rowconfigure(0, weight=1)
+        master.grid_rowconfigure(1, weight=2)
+        master.grid_columnconfigure(0, weight=1)
+        master.grid_columnconfigure(1, weight=1)
+
+        # creates a logo frame on top of the master frame
```
ssantichaivekin, line 23: I think it might look better to use capital letter to start a block comment.

empress_gui.py
```diff
@@ -0,0 +1,335 @@
...
+        self.view_cost_btn.grid(row=1, column=0)
+
+        # "Compute Reconciliations" button 
+        # displays reconciliation results(numbers) and three options(checkboxes) for viewing graphical analysis
+        self.compute_recon_button = tk.Button(self.func_frame, text="Compute Reconciliations", command=self.recon_analysis, state=tk.DISABLED)
+        self.compute_recon_button.grid(row=2, column=0)
+
+    def load_file(self):
+        """Loads in an input file and displays the number of leaves in each tree when "Load File" button is clicked."""
+        # creates a frame to display the number of leaves and the file path
```
ssantichaivekin, line 75: I think it might look better to use capital letter to start a block comment.

empress_gui.py
```diff
@@ -0,0 +1,335 @@
...
+                self.enable_recon_btn()
+            else:
+                messagebox.showinfo("Warning", "Please enter a non-negative number.")
+        except:
+            messagebox.showinfo("Warning", "Please enter a non-negative number.")
+    
+    def check_trans_entry(self, *args):
+        """Checks and updates the transfer cost when user enters a value in the entry box."""
+        try:
+            self.numbers_trans_entry = float(self.trans_cost.get())
```
ssantichaivekin, line 198: If you are using double box, you don't have to convert it to float I think.

empress_gui.py
```diff
@@ -0,0 +1,335 @@
...
+            self.valid_trans_entry = event.ydata
+            self.valid_loss_entry = event.xdata
+            self.enable_recon_btn()
+        else:
+            messagebox.showinfo("Warning", "Please click inside the axes bounds.")
+
+    def check_dup_entry(self, *args):
+        """Checks and updates the duplication cost when user enters a value in the entry box."""
+        try:
+            self.numbers_dup_entry = float(self.dup_cost.get())
```
ssantichaivekin, line 185: If you are using double variable/box, you don't have to convert it to float I think.

ssantichaivekin: I'm not sure though. Please consult the documentation.

empress_gui.py
```diff
@@ -0,0 +1,335 @@
...
+            if self.numbers_dup_entry >= 0:
+                self.valid_dup_entry = self.numbers_dup_entry
+                self.dup_cost.set(round(self.valid_dup_entry, 2))
+                self.enable_recon_btn()
+            else:
+                messagebox.showinfo("Warning", "Please enter a non-negative number.")
+        except:
```
ssantichaivekin, line 192: What exception are you trying to catch here? Please catch exact exceptions if possible:
```
except ValueError as e:
   # use e (the exception object) somehow
```

ssantichaivekin: https://docs.python.org/3/tutorial/errors.html

empress_gui.py
```diff
@@ -0,0 +1,335 @@
...
+            messagebox.showinfo("Warning", "Please enter a non-negative number.")
+    
+    def check_trans_entry(self, *args):
+        """Checks and updates the transfer cost when user enters a value in the entry box."""
+        try:
+            self.numbers_trans_entry = float(self.trans_cost.get())
+            if self.numbers_trans_entry >= 0:
```
ssantichaivekin, line 199: If you don't care about `numbers_trans_entry` afterwards, don't put it in `self`.

ssantichaivekin: I think you probably already know what's the difference between `var` and `self.var`, but if you aren't sure what the difference is, please read some documentation about python classes. https://www.askpython.com/python/python-self-variable https://docs.python.org/3/tutorial/classes.html

empress_gui.py
```diff
@@ -0,0 +1,335 @@
...
+                self.valid_trans_entry = self.numbers_trans_entry
+                self.trans_cost.set(round(self.valid_trans_entry, 2))
+                self.enable_recon_btn()
+            else:
+                messagebox.showinfo("Warning", "Please enter a non-negative number.")
+        except:
```
ssantichaivekin, line 205: Again, not clear what exception(s) you are trying to catch.

empress_gui.py
```diff
@@ -0,0 +1,335 @@
...
+        if self.valid_dup_entry is not None and self.valid_trans_entry is not None and self.valid_loss_entry is not None:
+            self.compute_recon_button.configure(state=tk.NORMAL)
+
+    def recon_analysis(self):
+        """Displays reconciliation results in numbers and further viewing options for graphical analysis."""
+        # creates a frame for the three checkbuttons
+        recon_checkbox_frame = tk.Frame(self.func_frame)
+        recon_checkbox_frame.grid(row=3, column=0, sticky="nsew")
+        recon_checkbox_frame.pack_propagate(False)
+        self.recon_space_btn_var = tk.IntVar()
```
ssantichaivekin, line 232: `BooleanVar()` maybe?

ssantichaivekin: I haven't read the code carefully so I might be wrong.

empress_gui.py
ssantichaivekin: Please look at what `DoubleVar.get` does. Does it raise an exception if it cannot be converted to float?

empress_gui.py
ssantichaivekin: If you are only trying to catch errors in the line
```python
self.numbers_trans_entry = float(self.trans_cost.get())
```
Then do it like this:
```python
try:
    self.numbers_trans_entry = float(self.trans_cost.get())
except ValueError as e:
    messagebox.showinfo(e)
```

empress_gui.py
```diff
@@ -0,0 +1,335 @@
...
+            if self.numbers_dup_entry >= 0:
+                self.valid_dup_entry = self.numbers_dup_entry
+                self.dup_cost.set(round(self.valid_dup_entry, 2))
+                self.enable_recon_btn()
+            else:
+                messagebox.showinfo("Warning", "Please enter a non-negative number.")
+        except:
+            messagebox.showinfo("Warning", "Please enter a non-negative number.")
+    
+    def check_trans_entry(self, *args):
```
ssantichaivekin, line 195: What does `*args` do?

empress_gui.py
```diff
@@ -0,0 +1,335 @@
...
+            if self.numbers_trans_entry >= 0:
+                self.valid_trans_entry = self.numbers_trans_entry
+                self.trans_cost.set(round(self.valid_trans_entry, 2))
+                self.enable_recon_btn()
+            else:
+                messagebox.showinfo("Warning", "Please enter a non-negative number.")
+        except:
+            messagebox.showinfo("Warning", "Please enter a non-negative number.")
+    
+    def check_loss_entry(self, *args):
```
ssantichaivekin, line 208: What does `*args` do?

empress_gui.py
```diff
@@ -0,0 +1,335 @@
...
+
+        # creates a logo frame on top of the master frame 
+        self.logo_frame = tk.Frame(master)
+        # sticky="nsew" means that self.logo_frame expands in all four directions (north, south, east and west) 
+        # to fully occupy the allocated space in the grid system (row 0 column 0&1)
+        self.logo_frame.grid(row=0, column=0, columnspan=2, sticky="nsew")
+        self.logo_frame.grid_propagate(False)
+
+        # adds background image
+        photo = tk.PhotoImage(file="./assets/Jane_TransLogo_thin.png")
```
ssantichaivekin, line 31: please use lowercase `jane_logo_thin.png`

empress_gui.py
Qing-Yang-1211: Done:)

empress_gui.py
Qing-Yang-1211: Done:)

empress_gui.py
Qing-Yang-1211: Done:)

empress_gui.py
Qing-Yang-1211: Done:)

empress_gui.py
Qing-Yang-1211: Done:)

empress_gui.py
Qing-Yang-1211: Done:)

empress_gui.py
Qing-Yang-1211: Done:)

empress_gui.py
Qing-Yang-1211: It raises an exception if the input is not float, e.g. " " or string input. I used self.numbers_trans_entry = float(self.trans_cost.get())
to capture this kind of exception.

Qing-Yang-1211: Yes, it does raise an exception if it cannot be converted to float.

empress_gui.py
Qing-Yang-1211: Yes, I don't need to convert it to float. Changed that. :))

empress_gui.py
Qing-Yang-1211: Ok! Done:)

empress_gui.py
Qing-Yang-1211: ok

empress_gui.py
Qing-Yang-1211: I am trying to catch string inputs, " " empty spaces and negative numbers. If I use except ValueError as e, then the messagebox.showinfo() line will not be executed.

empress_gui.py
Qing-Yang-1211: I think both are acceptable.

empress_gui.py
Qing-Yang-1211: The special syntax, *args and **kwargs in function definitions is used to pass a variable number of arguments to a function. Here, I think there are two arguments to be passed automatically but they are not needed in my code, so I used *args. This is the error if *args is not there:
Exception in Tkinter callback
Traceback (most recent call last):
  File "/Applications/anaconda3/lib/python3.7/tkinter/__init__.py", line 1705, in __call__
    return self.func(*args)
TypeError: check_dup_entry() takes 1 positional argument but 2 were given

empress_gui.py
Qing-Yang-1211: See above.

empress_gui.py
Qing-Yang-1211: Done:)

Qing-Yang-1211:
> It seems that the image is not in pull request. Please exclude asset images in `.gitignore` by adding
> 
> ```
> # Exclude assets
> !/assets
> ```
> 
> line at the bottom of the file.
> 
> If you want to know what `.gitignore` is, feel free to look it up. See the first and second answers in this post: https://stackoverflow.com/questions/32504123/unignore-specific-files-in-subdirectory-with-gitignore

Done:)

empress_gui.py
```diff
@@ -0,0 +1,369 @@
...
+            self.loss_input.set(round(event.xdata, 2))
+            # Enables the next step, viewing reconciliation result
+            self.dup_cost = 1.00
+            self.trans_cost = event.ydata
+            self.loss_cost = event.xdata
+            self.update_recon_btn()
+        else:
+            messagebox.showinfo("Warning", "Please click inside the axes bounds.")
+
+    def update_recon_btn(self):
```
ssantichaivekin, line 248: I think we should move this up closer to `validate_loss_input`

empress_gui.py
```diff
@@ -0,0 +1,369 @@
...
+            self.histogram_window = tk.Toplevel(self.master)
+            self.histogram_window.geometry("400x400")
+            self.histogram_window.title("View p-value histogram")
+            HistogramWindow(self.histogram_window)
+        if self.histogram_btn_var.get() == False:
+            if self.histogram_window.winfo_exists() == 1:
+                self.histogram_window.destroy()
+
+# View reconciliation space 
+class ReconSpaceWindow:
```
ssantichaivekin, line 334: maybe have this class derive from `tk.Frame`

empress_gui.py
```diff
@@ -0,0 +1,369 @@
...
+    def __init__(self, master):
+        self.master = master        
+        self.frame = tk.Frame(master)
+        self.frame.pack(fill=tk.BOTH, expand=1)
+        self.frame.pack_propagate(False)
+
+# View reconciliations 
+class ReconsWindow:
```
ssantichaivekin, line 342: maybe have this class derive from `tk.Frame`

PR #61, opened 2020-06-06
Clean up docstrings and remove dead code

rlhadas:
Greedy.py is not used so it was removed.  Several other files had functions that are no longer used and were removed.

PR #66, opened 2020-06-09
Clean up docstrings and remove dead code

ssantichaivekin:
From Prof Ran: Greedy.py is not used so it was removed. Several other files had functions that are no longer used and were removed.

The pull request is the same as #61 . I don't know why, but the files changed list on that pull request got buggy, so I create a new pull request.


ssantichaivekin:
I have already tested this in #61 , so I will just merge.

PR #67, opened 2020-06-10
Refactor subcommands to own files

wutron:
Update `empress.py` and add new folder `commands`.


ssantichaivekin:
Thanks Prof Wu @wutron a lot for doing this :) I agree that the cli should be separated into four files. However, I think the solution proposed in this pull request is too hacky. It involves using the python import for a purpose that it's generally not used for. It also uses `sys.argv` alongside with the `argparse` library, which I find to be quite odd.  I took a lot of design inspiration from this pull request and wrote another pull request #123 that purely uses `argparse` library without using python import and `sys.argv`. I will close this pull request in favor of #123.

PR #73, opened 2020-06-12
Slight cleanup to the clustering CLI

aremath:
Cleaned up the clustering CLI slightly by making some arguments clearer. Removed the default for the number of clusters: this is probably something the user should know in advance, or at least should be asked to specify.


ssantichaivekin:
Thanks for fixing the cli! Could you also update `./github/workflow/python_app.yaml`? There's a line in there that says `pipenv run python empress_cli.py -fn examples/heliconius.newick clumpr --median --nmprs 4 --support` which should be updated along with the readme :)

PR #74, opened 2020-06-14
Recon vis

rlhadas:
This is intended to replace both the topo_sort and vis_trans branches.  All of the functionality of those branches has been included here.  The current rendering is correct but not pretty.  Justin and Trenton will be working on making the drawing much nicer.


ssantichaivekin:
Pull request abandoned. (Prof Ran, Justin, Trenton will create a new one).

PR #78, opened 2020-06-16
Update recon-vis branch

ssantichaivekin:
Fix imports.
Use updated Rose's code.
Refactor visualization: move `fig.show()` outside.
Move examples to test folder.
Fix imports in Roses's unittests.
Update current print tests to unittests.


test/clumpr/recon_vis/test_Tree.py
```diff
@@ -0,0 +1,62 @@
...
+            ('m9', 'rosina_WestPA'): ('m9', 'rosina_WestPA', None, None),
+            ('m8', 'm11'): ('m8', 'm11', ('m11', 'melpomene_EastC'), ('m11', 'cythera_WestE')),
+            ('m11', 'melpomene_EastC'): ('m11', 'melpomene_EastC', None, None),
+            ('m11', 'cythera_WestE'): ('m11', 'cythera_WestE', None, None)}
+
+    def test_convert(self):
+        host_converted = dict_to_tree(self.host, tree.TreeType.HOST)
+        self.assertTrue(isinstance(host_converted, tree.Tree))
+
+    def test_leaf_list(self):
```
ssantichaivekin, line 45: I think this test will be easier to write/understand if we use a smaller tree. If the tree is small, we can write out the expected `leaves` by hand and just check `self.assertEqual(leaves, expected)`.

test/clumpr/recon_vis/test_Tree.py
```diff
@@ -0,0 +1,62 @@
...
+        host_converted = dict_to_tree(self.host, tree.TreeType.HOST)
+        leaves = host_converted.leaf_list()
+        for leaf in leaves:
+            self.assertTrue(isinstance(leaf, tree.Node))
+            self.assertTrue(leaf.is_leaf())
+        self.assertEqual(len(leaves), 12)
+
+    def test_postorder_list(self):
```
ssantichaivekin, line 53: Same here.

test/clumpr/recon_vis/test_recon_viewer.py
```diff
@@ -0,0 +1,68 @@
...
+    recon_dict2 = {('n0', 'm1'): [('T', ('n1', 'm1'), ('n4', 'm4'))],
+                   ('n1', 'm1'): [('D', ('n2', 'm1'), ('n3', 'm1'))],
+                   ('n2', 'm1'): [('L', ('n2', 'm2'), (None, None))],
+                   ('n3', 'm1'): [('L', ('n3', 'm3'), (None, None))],
+                   ('n2', 'm2'): [('C', (None, None), (None, None))],
+                   ('n3', 'm3'): [('C', (None, None), (None, None))],
+                   ('n4', 'm4'): [('C', (None, None), (None, None))]}
+
+    def test_render_1(self):
+        fig = recon_viewer.render(self.host_dict1, self.parasite_dict1, self.recon_dict1,
```
ssantichaivekin, line 52: The idea here is that if someone wants to look at how a code work, they can go to the test folder and run the tests and step through them in a debugger.

test/topo_sort/test_Tree.py
```diff
@@ -1,41 +0,0 @@
-# test_Tree.py
```
ssantichaivekin, line 1: Moved.

test/clumpr/recon_vis/test_Tree.py
```diff
@@ -0,0 +1,62 @@
...
+        host_converted = dict_to_tree(self.host, tree.TreeType.HOST)
+        nodes = host_converted.postorder_list()
+        for node in nodes:
+            self.assertTrue(isinstance(node, tree.Node))
+        self.assertEqual(len(nodes), len(self.host))
+
```
ssantichaivekin, line 59: I think we could add more tests for the tree type.

test/clumpr/recon_vis/test_Tree.py
```diff
@@ -0,0 +1,62 @@
+# test_Tree.py
+# Tests the Tree class and tree_format_converter.py
+import unittest
+
+from empress.clumpr.recon_vis import tree
+from empress.clumpr.recon_vis.utils import dict_to_tree
```
ssantichaivekin, line 6: Because `recon_vis.tree` is not dependent on `recon_vis.utils`, `test_Tree` should avoid importing `recon_vis.utils` as this can make the test confusing to read.

test/clumpr/recon_vis/test_Tree.py
```diff
@@ -0,0 +1,62 @@
+# test_Tree.py
+# Tests the Tree class and tree_format_converter.py
```
ssantichaivekin, line 2: Moved from `test/topo_sort/test_Tree.py` and converted to unittest.

empress/clumpr/recon_vis/tree.py
```diff
@@ -19,14 +19,10 @@ def __init__(self, name):
         self.parent_node = None     # Node:  parent Node or None
         self.layout = None          # NodeLayout object: layout of this node
 
-    # The @property decorator allows this to be called as .is_leaf rather than .is_leaf()
-    @property
```
ssantichaivekin, line 23: I don't think this should be a property because it's not a way of accessing an instance variable. For example, `string.isdigit()`, `string.islower()`, `string.upper()` in are not properties.

ssantichaivekin: If there is a variable named `_is_leaf`, then it would make sense to make `is_leaf` a property.

empress/clumpr/recon_vis/utils.py
```diff
@@ -22,55 +22,11 @@ def convert_to_objects(host_dict, parasite_dict, recon_dict):
     :return - corresondoing host_tree and parasite_tree Tree objects
         and recon Reconciliation object
     """
-    host_tree, parasite_tree = build_trees_with_temporal_order(host_dict, parasite_dict, recon_dict)
+    host_tree, parasite_tree, ok = build_trees_with_temporal_order(host_dict, parasite_dict, recon_dict)
```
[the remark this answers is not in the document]
rlhadas: Here's what we need to do...  Rose will modify the temporal ordering code so that if the current (strict) temporal ordering isn't valid, she will try to find a second (weak) temporal ordering.  Then, the build_trees_with_temporal order should return an indication if the ordering was strict, weak, or no ordering.  Strict and weak are both fine for rendering.  If there was no ordering, the gui/cli will indicate that.

test/clumpr/recon_vis/test_Tree.py
```diff
@@ -0,0 +1,62 @@
...
+            ('m8', 'm9'): ('m8', 'm9', ('m9', 'm10'), ('m9', 'rosina_WestPA')),
+            ('m9', 'm10'): ('m9', 'm10', ('m10', 'melpomene_WestPA'), ('m10', 'rosina_WestCR')),
+            ('m10', 'melpomene_WestPA'): ('m10', 'melpomene_WestPA', None, None),
+            ('m10', 'rosina_WestCR'): ('m10', 'rosina_WestCR', None, None),
+            ('m9', 'rosina_WestPA'): ('m9', 'rosina_WestPA', None, None),
+            ('m8', 'm11'): ('m8', 'm11', ('m11', 'melpomene_EastC'), ('m11', 'cythera_WestE')),
+            ('m11', 'melpomene_EastC'): ('m11', 'melpomene_EastC', None, None),
+            ('m11', 'cythera_WestE'): ('m11', 'cythera_WestE', None, None)}
+
+    def test_convert(self):
```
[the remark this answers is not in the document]
rlhadas: OK.

empress/clumpr/recon_vis/utils.py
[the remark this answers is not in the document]
ssantichaivekin: Currently, it seems that Rose's code responds `True` when there is a strict ordering and `False` if there is no strict ordering (and set the host and parasite as none). Do you think the current code should respond to that signal (here it is indicated as ok)?

PR #82, opened 2020-06-17
Split clumpr -> cluster, histogram, reconcile

aremath:
Split clumpr into high-level packages clustering, histogram, and reconcile.


ssantichaivekin:
Looks good to me so far, did you look for all the references and change their imports too?

ssantichaivekin:
Does the gui still runs correctly?

PR #87, opened 2020-06-18
Modify build_trees_with_temporal_order to return ConsistencyType

JingyiRose:
- modified `recon_vis/utils` `convert_to_objects` function to additionally return a `ConsistencyType` that indicates whether a given reconciliation is strongly consistent, weakly consistent, or not consistent

- modified the corresponding `test_utils`

### DIFF
--- a/empress/reconcile/recon_vis/recon_viewer.py
+++ b/empress/reconcile/recon_vis/recon_viewer.py
@@ -17,12 +17,15 @@ def render(host_dict, parasite_dict, recon_dict, show_internal_labels=False, sho
     :param parasite_dict:  Parasite tree represented in dictionary format
     :param recon_dict: Reconciliation represented in dictionary format
     """
-    # TODO: consistency_type is not used
-    host_tree, parasite_tree, recon, consistency_type = utils.convert_to_objects(host_dict, parasite_dict, recon_dict)
-    fig = plot_tools.FigureWrapper("Reconciliation")
-    render_host(fig, host_tree, show_internal_labels)
-    host_lookup = host_tree.name_to_node_dict()
-    render_parasite(fig, parasite_tree, recon, host_lookup, show_internal_labels, show_freq)
+    # convert host and parasite dicts to objects and populate nodes with their temporal order
+    host_tree, parasite_tree, consistency_type = utils.build_trees_with_temporal_order(host_dict, parasite_dict, recon_dict)
+    recon = utils.dict_to_reconciliation(recon_dict)
+    
+    fig = plot_tools.FigureWrapper(consistency_type)
+    if consistency_type != utils.ConsistencyType.NO_CONSISTENCY:
+        render_host(fig, host_tree, show_internal_labels)
+        host_lookup = host_tree.name_to_node_dict()
+        render_parasite(fig, parasite_tree, recon, host_lookup, show_internal_labels, show_freq)
     return fig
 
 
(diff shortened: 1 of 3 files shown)


empress/reconcile/recon_vis/utils.py
```diff
@@ -172,31 +179,43 @@ def dict_to_recongraph(old_recon_graph: Dict[Tuple, List]):
 
 # Temporal ordering utilities
 
-def build_trees_with_temporal_order(host_tree: dict, parasite_tree: dict, reconciliation: dict) \
-        -> Tuple[tree.Tree, tree.Tree, bool]:
+def build_trees_with_temporal_order(host_tree: dict, parasite_tree: dict, reconciliation: dict,) \
```
ssantichaivekin, line 182: Is there a reason you added the extra `,`? It wouldn't give you syntax error but I just wonder whether it's a typo.

empress/reconcile/recon_vis/recon_viewer.py
```diff
@@ -17,7 +17,8 @@ def render(host_dict, parasite_dict, recon_dict, show_internal_labels=False, sho
     :param parasite_dict:  Parasite tree represented in dictionary format
     :param recon_dict: Reconciliation represented in dictionary format
     """
-    host_tree, parasite_tree, recon = utils.convert_to_objects(host_dict, parasite_dict, recon_dict)
+    # TODO: consistency_type is not used
```
ssantichaivekin, line 20: I feel like we should not put todos in the code because it can easily be forgotten. I think what we can do is either create an issue and assign Prof Ran to the issue and then post the link to the issue under the todo, or remove this `# TODO:`.

empress/reconcile/recon_vis/utils.py
```diff
@@ -9,22 +9,29 @@
...
     :param host_dict - dictionary representation of host tree
     :param parasite_dict - dictionary representation of parasite tree
     :param recon_dict - dictionary representation of reconciliation
-    :return - corresondoing host_tree and parasite_tree Tree objects
-        and recon Reconciliation object
+    :return - corresondoing host_tree and parasite_tree Tree objects,
+        recon Reconciliation object and a ConsistencyType
     """
-    host_tree, parasite_tree, ok = build_trees_with_temporal_order(host_dict, parasite_dict, recon_dict)
+    host_tree, parasite_tree, consistency_type = build_trees_with_temporal_order(host_dict, parasite_dict, recon_dict)
```
ssantichaivekin, line 32: I think this function should be removed. We don't want to modify this function every time we modify `build_trees_with_temporal_order`. I think the user could run the two lines themselves. Would you mind looking for the occurrences of `convert_to_objects` and change it to `build_trees_with_temporal_order` and `dict_to_reconciliation`?

empress/reconcile/recon_vis/utils.py
```diff
@@ -265,11 +284,14 @@ def uniquify(elements):
     """
     return list(set(elements))
 
-def build_temporal_graph(host_tree, parasite_tree, reconciliation):
+def build_temporal_graph(host_tree, parasite_tree, reconciliation, add_strong_constraints = True):
```
ssantichaivekin, line 287: I think the name `use_strong_constraints` might be more indicative. This is very minor though. Feel free to keep as is.

ssantichaivekin: Should the `constraints` be plural or singular `constraint`?

JingyiRose: when we use strong constraints, we basically add two more constraints for every transfer event, so I think add_strong_constraints is more indicative.

Constraints are plural to be consistent with the comments in the code

PR #88, opened 2020-06-18
Modify ReconInput to read from two/three files

zachwitz:
Added implicit matching to newickTreeReader


empress/newickTreeReader.py
```diff
@@ -0,0 +1,215 @@
...
+        else:
+            return None
+
+        map_string = map_file.read()
+        map_list = map_string.split()
+        self.phi = ReconInput.parse_phi(map_list)
+        map_file.close()
+
+    @staticmethod
+    def imply_mapping(host_tree : tree, parasite_tree : tree):
```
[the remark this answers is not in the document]
ssantichaivekin: I believe the `host_tree` and `parasite_tree` type is `dict`.

empress/newickTreeReader.py
```diff
@@ -0,0 +1,215 @@
...
+        """
```
ssantichaivekin, line 58: It seems like this function can produce an invalid mapping if some of names follow the convention while some of the names doesn't follow the convention. If the inferred mapping is invalid, I think we should return `None` instead of the invalid mapping.

empress/fileInput.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def __init__(self, host_tree = None, host_distances = None, parasite_tree = None, 
+                parasite_distances = None, phi = None):
+        self.host_tree = host_tree
+        self.host_distances = host_distances
+        self.parasite_tree = parasite_tree
+        self.parasite_distances = parasite_distances
+        self.phi = phi
+
+    def complete(self, host_file: str, parasite_file: str, mapping_file = None):
```
ssantichaivekin, line 21: Maybe rename the function to `read(...)`.

empress/fileInput.py
```diff
@@ -0,0 +1,220 @@
...
+
+class ReconInput(object):
+    """
+    Storage class for the newick data (trees, tip mapping, and optional distance parameters)
+    Distances encode branch lengths in the newick file, if given
+    """
+
+    def __init__(self, host_tree = None, host_distances = None, parasite_tree = None, 
+                parasite_distances = None, phi = None):
+        self.host_tree = host_tree
```
ssantichaivekin, line 15: After you write the function that verifies whether the tree and the mapping is correct, it would be nice if you add the verification here too. :)

empress/fileInput.py
```diff
@@ -0,0 +1,220 @@
...
+
+    @staticmethod
+    def read_newick_tree(file_name: str, tree_type: str):
+        """
+        Take a filename of a newick tree as input and returns the parsed version of that tree.
+        :param filename <str>
+        :param tree_type (either "host" or "parasite")
+        """
+        if isinstance(file_name, str):
+            host_file = open(file_name, 'r')
```
ssantichaivekin, line 107: This shouldn't be called `host_file`.

empress/fileInput.py
zachwitz: I don't know this is just what Prof Ran told me to name it.

ssantichaivekin:
Hi Zach, I will close this pull request because the issue is resolved in #100 . Feel free to open a new pull request if you want to add tip mapping inference.

PR #91, opened 2020-06-19
Add visualization tests and upload output as build artifacts

### LINKED ISSUE #83: Create "tests" for visualization that are not unittests
Create automated "tests" for visualizations that run every time a pull request is created to master. The test does not actually test the validity of the output images. It just put the output images as build artifacts of the pull request.

ssantichaivekin:
Add visualization examples, set up ci to automatically run the examples and put the output in the build artifacts.

Resolves #83 .

---

Previously this week, Prof Ran told me that the tests I wrote for visualization code might be too hard for people to understand and navigate. After thinking about it, I agree that it is quite odd that a person has to pause in the debugger to see the output of a test. Still, I want the test to be automated. I was thinking about a solution that would both be easy to understand and is also automated.

I think a solution to this is, instead of using unittests, we can write simple scripts for testing visualization code that will write an output to a file. The scripts will be similar to the scripts Prof Ran originally wrote for visualization testing. These scripts will all be in a folder named `draw` at top-level. The ci will call each script in this folder and will upload the images drawn by the script to what's called the "artifact" of the build as shown below:

Screen Shot 2020-06-19 at 1 24 57 PM
Screen Shot 2020-06-19 at 1 25 01 PM
Screen Shot 2020-06-19 at 1 25 14 PM
Screen Shot 2020-06-19 at 1 26 41 PM


draw/draw_empress_wrapper.py
```diff
@@ -0,0 +1,136 @@
...
+# Run this file to generate drawing examples
+
+import empress
+import matplotlib.pyplot as plt
+from pathlib import Path
+from draw import common
+
+common.create_output_folder()
+
+example_input_path = "./examples/test-size5-no924.newick"
```
rlhadas, line 11: I suggest that we move away from .newick files.  This is the "old" input format.  We're now moving to having three input files comprising a host tree (.nwk), parasite tree (.nwk), and mapping (.mapping).  It's easy to create these from a .newick file:  The first tree in the .newick file goes in the host's .nwk file, the second one goes in the parasite's .nwk file, and the remaining stuff goes in the .mapping file.  Since this is the way that we plan to request user input, it makes sense for our tests to use this too.

ssantichaivekin: Yes. We should change this in Zach's pull request which will change the one file input to three files input.

Issue #28, opened 2020-05-26
Add Input Sanitization


ssantichaivekin:
Add more input sanitization (such as checking whether the tree and the tip mapping is valid) in the code.

PR #101, opened 2020-06-24
Modify ReconInput to use three files instead of one

### LINKED ISSUE #100: Modify automated newick trees generation to use two/tree files instead of one
Will disable the current unittest that uses automated tree generation

ssantichaivekin:
- The cli now has three file arguments `--host` `--parasite` and `--mapping` instead of the original `-fn/--filename`. Currently, they are all required.
- Two ways to construct `ReconInput` class: We can use the constructor `ReconInput.from_files(host_fname, parasite_fname, mapping_fname)` or we can create an empty `ReconInput` class via init and then call `read_host(fname)` `read_parasite(fname)` and then `read_mapping(fname)`.
- If reading the input fails at any point, it raises the exception of type `empress.input_reader.ReconInputError` with appropriate error message. If the user is unfamiliar with python exceptions, feel free to read the docs here: https://docs.python.org/3/tutorial/errors.html
- Updated examples and automated tests
- There is one unit test that relies on newick files generation. I haven't modified the unittest and the files generation. The unittest is currently skipped.
- Add tip mapping verification. All constructors of `ReconInput` call this verification directly or indirectly. The verification throws an exception if the tip mapping is invalid.
- Add tests to `test/test_input_reader.py`

Resolves #100


ssantichaivekin:
Let's do that in another pull request. I will open an issue for this.

ssantichaivekin:
Resolves #28 .

PR #102, opened 2020-06-25
Modify reconcile to compute ReconGraph event frequencies

JingyiRose:
1. add `event_scores` in `ReconGraphWrapper` and `ReconciliationWrapper`
2. compute `event_scores` for `ReconGraphWrapper` in `__init__` function
2. modify `dict_to_reconciliation` in `recon_vis/utils.py` to take in an `event_scores` argument and populate frequency for each event in the `Reconciliation` object it creates
3. change gene/species notation to parasite/host in `empress.__init__.py`
4. add an example for rendering reconciliation with event frequency in `draw/draw_recon_viewer.py`


empress/recon_vis/recon_viewer.py
```diff
@@ -12,7 +12,7 @@
     PARASITE_EDGE_COLOR, VERTICAL_OFFSET, COSPECIATION_OFFSET
 
 
-def render(host_dict, parasite_dict, recon_dict, show_internal_labels=False,
+def render(host_dict, parasite_dict, recon_dict, event_scores = None, show_internal_labels=False,
```
JingyiRose, line 15: should event_scores be an optional argument?

empress/recon_vis/utils.py
```diff
@@ -87,7 +87,7 @@ def _find_roots(old_recon_graph) -> List[MappingNode]:
             roots.append(mapping)
     return roots
 
-def dict_to_reconciliation(old_recon: Dict[Tuple, List]):
+def dict_to_reconciliation(old_recon: Dict[Tuple, List], event_scores = None):
```
ssantichaivekin, line 90: You might also want to modify `dict_to_recongraph(...)` to also receive `event_scores`

JingyiRose: does this still exist?

empress/__init__.py
```diff
@@ -107,17 +108,18 @@ def __init__(self, recongraph: dict, roots: list, n_recon: int, recon_input: Rec
         self.total_cost = total_cost
         self.n_recon = n_recon
         self.roots = roots
+        self.event_scores = self.__compute_event_frequencies()
```
ssantichaivekin, line 111: It seems that the code implicitly recompute event scores when you cluster, which is not the expected behavior.

empress/__init__.py
```diff
@@ -145,19 +147,19 @@ def median(self) -> ReconciliationWrapper:
...
 
         med_counts_dict = median.get_med_counts(median_reconciliation, roots_for_median)
 
         random_median = median.choose_random_median_wrapper(median_reconciliation, roots_for_median, med_counts_dict)
         median_root = _find_roots(random_median)[0]
         return ReconciliationWrapper(random_median, median_root, self.recon_input, self.dup_cost, self.trans_cost,
-                                     self.loss_cost, self.total_cost)
+                                     self.loss_cost, self.total_cost, self.event_scores)
 
     def cluster(self, n) -> List['ReconGraphWrapper']:
```
ssantichaivekin, line 164: It seems that the code implicitly recompute event scores when you cluster, which is not the expected behavior.

ssantichaivekin: You might consider adding tests that make sure the event freq stays the same after you cluster or median, but I don't want to take more of your time.

empress/__init__.py
ssantichaivekin: What you could do is to add `event_scores` as a required argument to `__init__` on line 102. Then, we could modify `_compute_event_frequencies` to be a static method which takes `ReconInput` and `recon_dict: dict`. We only call `_compute_event_frequencies` when we call `reconcile`. At other points, we copy the event scores down to lower level.

JingyiRose: changed to use set_event_frequencies in reconcile

empress/__init__.py
```diff
@@ -181,6 +183,19 @@ def cluster(self, n) -> List['ReconGraphWrapper']:
                                   self.total_cost))
         return new_graphs
 
+    def __compute_event_frequencies(self) -> dict:
```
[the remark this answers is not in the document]
JingyiRose: changed to a public function

empress/__init__.py
```diff
@@ -166,21 +168,34 @@ def cluster(self, n) -> List['ReconGraphWrapper']:
...
         for graph in graphs:
             roots = _find_roots(graph)
             n = recongraph_tools.count_mprs_wrapper(roots, graph)
             new_graphs.append(
                 ReconGraphWrapper(graph, roots, n, self.recon_input, self.dup_cost, self.trans_cost, self.loss_cost,
-                                  self.total_cost))
+                                  self.total_cost, self.event_scores))
         return new_graphs
 
+    def set_event_frequencies(self) -> dict:
```
ssantichaivekin, line 186: It does not return a dict.

PR #108, opened 2020-07-01
Rewrite ReconInput generator

### LINKED ISSUE #100: Modify automated newick trees generation to use two/tree files instead of one

ssantichaivekin:
Rewrite the generator that generates all or random `ReconInput` objects. Add saving functionality to `ReconInput` so that one can save generated inputs.

Resolves #100 .


empress/miscs/input_generator.py
```diff
@@ -0,0 +1,161 @@
...
+    """
+    We don't actually generate all possible trees in this
+    format. In the code, you will see that we only generate a right-heavy
+    tree, meaning that the right subtree has equal or greater number of
+    nodes than the left subtree. We do this to reduce the number of trees
+    that are isomorphic to one another.
+    """
+    tree_dict = {}
+    if n_leaves == 1:
+        tree_dict[top_name] = ('Top', prefix + '0', None, None)
```
JingyiRose, line 44: can we handle this case in generate_all_trees_helper by checking whether n_leaves is 0?

ssantichaivekin: I'm not sure I understand your comment. We only call the `_generate_all_trees_helper` if we have at least two leaves. Are you saying that even if we have one leaf, we should call both `_generate_all_trees_helper` on the left subtree and on the right subtree even if there is only one subtree? Would you mind clarifying how this should be modified?

ssantichaivekin:
I'm thinking that `miscs` could contain things that are not relevant to the main workflow of empress. I'm thinking of a name that represents "other things that are not very important".

empress/miscs/input_generator.py
JingyiRose: yeah, your understanding is correct. This is only a minor suggestion, the code works fine either way~

ssantichaivekin: I still don't understand your comment, and it does not seem that you are planning to explain it, so I will just merge.

PR #110, opened 2020-07-02
Rename event_scores to event_frequencies

### LINKED ISSUE #109: Rename event_scores to event_frequencies
Rename the dictionary that stores the event frequencies from `event_scores` to `event_frequencies`.

ssantichaivekin:
Resolves #109 .


empress/reconcile/median.py
```diff
@@ -68,14 +68,14 @@ def mapping_node_sort(ordered_gene_node_list, ordered_species_node_list, mapping
     return sorted_list
 
 
-def generate_scores(preorder_mapping_node_list, dtl_recon_graph, gene_root, normalize=True):
+def generate_frequencies_dict(preorder_mapping_node_list, dtl_recon_graph, gene_root, normalize=True):
     """
     Computes frequencies for every event
     :param preorder_mapping_node_list: A list of all mapping nodes in DTLReconGraph in double preorder
     :param dtl_recon_graph: The DTL reconciliation graph that we are scoring
```
JingyiRose, line 75: can you edit this comment to say sth like ":param dtl_recon_graph: The DTL reconciliation graph whose events we want to annotate with their corresponding frequencies"

empress/reconcile/median.py
```diff
@@ -168,42 +168,42 @@ def count_mprs(mapping_node, dtl_recon_graph, counts):
...
 
-def calculate_scores_for_children(mapping_node, dtl_recon_graph, event_scores, mapping_scores, counts):
+def calculate_event_frequencies_for_children(mapping_node, dtl_recon_graph, event_frequencies, node_frequencies, counts):
     """
-    This function calculates the frequency score for every mapping node that is a child of an event node that is a
-    child of the given mapping node, and stores them in scoredGraph.
-    :param mapping_node: The mapping node that is the parent of the two scores we will compute
+    This function calculates the frequency for every mapping node that is a child of an event node that is a
+    child of the given mapping node, and stores them in dtl_recon_graph.
+    :param mapping_node: The mapping node that is the parent of the two frequencies we will compute
```
JingyiRose, line 175: rephrase it to:   :param mapping_node: The mapping node that is the parent of the two mapping nodes whose frequencies we will compute

empress/reconcile/median.py
```diff
@@ -93,36 +93,36 @@ def generate_scores(preorder_mapping_node_list, dtl_recon_graph, gene_root, norm
...
     for mapping_node in preorder_mapping_node_list:
-        scores[mapping_node] = 0.0
+        node_frequencies[mapping_node] = 0.0
 
     # This entry is going to be thrown away, but it seems neater to just let calculateScoresOfChildren
     # add scores to an unused entry than to check to see if they are (None, None) in the first place.
-    scores[(None, None)] = 0.0
+    node_frequencies[(None, None)] = 0.0
 
     # The scored graph is like the DTLReconGraph, except instead of individual events being in a list, they are the
```
JingyiRose, line 105: Same as the comment above, I think we are not actually explicitly constructing a "scored graph" with mapping nodes as keys and events and their frequencies as values, so this comment can be misleading

empress/reconcile/median.py
```diff
@@ -68,14 +68,14 @@ def mapping_node_sort(ordered_gene_node_list, ordered_species_node_list, mapping
...
 
 
-def generate_scores(preorder_mapping_node_list, dtl_recon_graph, gene_root, normalize=True):
+def generate_frequencies_dict(preorder_mapping_node_list, dtl_recon_graph, gene_root, normalize=True):
     """
     Computes frequencies for every event
     :param preorder_mapping_node_list: A list of all mapping nodes in DTLReconGraph in double preorder
     :param dtl_recon_graph: The DTL reconciliation graph that we are scoring
     :param gene_root: The root of the gene tree
     :return: 0. A file structured like the DTLReconGraph, but with the lists of events replaced
```
JingyiRose, line 77: I think we are not actually explicitly constructing a "scored graph" with mapping nodes as keys and events and their frequencies as values, so this comment can be misleading

PR #112, opened 2020-07-02
Add event counting to empress.ReconciliationWrapper

ssantichaivekin:
```python
cospec_count, dup_count, trans_count, loss_count = reconciliation.count_events()
```

### DIFF
--- a/empress/__init__.py
+++ b/empress/__init__.py
@@ -95,23 +95,26 @@ class ReconciliationWrapper(Drawable):
         recon_viewer.render(self.recon_input.host_dict, self.recon_input.parasite_dict, self._reconciliation,
                             axes=axes)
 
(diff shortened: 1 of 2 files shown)


empress/__init__.py
```diff
@@ -95,6 +95,25 @@ def draw_on(self, axes: plt.Axes):
         recon_viewer.render(self.recon_input.host_dict, self.recon_input.parasite_dict, self._reconciliation,
                             axes=axes)
 
+    def count_events(self) -> Tuple[int, int, int]:
```
rlhadas, line 98: We should also count (and, ultimately report) the number of cospeciations.  The truth is that we can infer the number of cospeciations from dup_count, trans_count, and loss_count, but we may as well compute them here just as we do with other events.

ssantichaivekin: ok

PR #121, opened 2020-07-08
Add more detailed error messages for input_reader

ssantichaivekin:


ssantichaivekin:
I have added more information for the tip mapping error. For the host/parasite file, I think that's the best I could do. I think if the data is malformed like that, the `Phylo.read(StringIO(newick_string), "newick")` should have thrown an exception. However, `Phylo.read(...)` does not do that but just give us back weird data which breaks our code when we use it.

ssantichaivekin:
On a side note, I think it's okay as long as it says "cannot read host tree," "cannot read parasite tree," or "cannot read tip mapping" as part of the error message.

PR #122, opened 2020-07-08
GUI: break up three functions

### LINKED ISSUE #119: Break up functions in gui
See todos in the current gui code.

Qing-Yang-1211:
Break up the __init__ function into:
init_frames(self, master)
init_load_files(self)
init_view_tanglegram(self)
init_view_cost_space(self)
init_compute_reconciliations(self)
init_view_solution_space(self)
init_view_reconciliations(self)
init_view_pvalue_histogram(self)
init_windows(self)

Break up refresh_when_new_input_files_loaded into:
refresh_when_reload_host
refresh_when_reload_parasite
refresh_when_reload_mapping
reset

Break up update_input_files_info into:
update_host_info
update_parasite_info
update_mapping_info

Resolves #119


Qing-Yang-1211:
For the breaking up of the refresh_when_new_input_files_loaded function, I didn't put the reset function only at the place where the mapping file is reloaded because if the user reloads in the host/parasite file first, some widgets will remain on the gui and will cause errors if they are clicked. I used the variable self.first_time_loading_files to determine if the user is loading in all three files for the first time(first round of loading input files), and the variable self.need_to_reset to only reset when the user reloads the first new input file, so that the reset function will only be run once for each round of reloading files and it will not be run at the first round of loading input files.

empress_gui.py
```diff
@@ -160,69 +140,126 @@ def __init__(self, master):
...
         self.cost_space_window = None
         self.entire_space_window = None
         self.set_num_cluster_window = None
         self.one_MPR_window = None
         self.view_solution_space_window = None
         self.view_reconciliations_window = None
         self.view_pvalue_histogram_window = None
+    
+    def refresh_when_reload_host(self):
+        if self.first_time_loading_files == True:
```
ssantichaivekin, line 202: remove `== True`

empress_gui.py
```diff
@@ -160,69 +140,126 @@ def __init__(self, master):
...
+            self.recon_input.read_host(self.host_file_path)
+            self.host_tree_info.destroy()
+        else:
+            if self.need_to_reset == True:
```
ssantichaivekin, line 206: remove `== True`

empress_gui.py
```diff
@@ -160,69 +140,126 @@ def __init__(self, master):
...
+                self.parasite_tree_info.destroy()
+                self.mapping_info.destroy()
+            else:
+                self.recon_input.read_host(self.host_file_path)
+                self.host_tree_info.destroy()
+                self.parasite_tree_info.destroy()
+                self.mapping_info.destroy()
+    
+    def refresh_when_reload_parasite(self):
+        if self.first_time_loading_files == True:
```
ssantichaivekin, line 219: remove `== True`

empress_gui.py
```diff
@@ -160,69 +140,126 @@ def __init__(self, master):
...
+                self.parasite_tree_info.destroy()
+                self.mapping_info.destroy()
+            else:
+                self.recon_input.read_host(self.host_file_path)
+                self.recon_input.read_parasite(self.parasite_file_path)
+                self.parasite_tree_info.destroy()
+                self.mapping_info.destroy()
+    
+    def refresh_when_reload_mapping(self):
+        if self.first_time_loading_files == True:
```
ssantichaivekin, line 236: remove `== True`

empress_gui.py
ssantichaivekin: Use `if boolean_variable:` or `if not boolean_variable:` in favor of `if boolean_variable == True:` or `if boolean_variable == False:`.

empress_gui.py
ssantichaivekin: It seems that this code block
```python
self.recon_input.read_host(self.host_file_path)
self.host_tree_info.destroy()
self.parasite_tree_info.destroy()
self.mapping_info.destroy()
```
is shared by both the if and else cases. It might be nice to move it out of if/else. See example below:
```python
if x:
    do_a()
    do_b()
else:
    do_b()
```
should be changed to
```python
if x:
    do_a()
do_b()
```
Please do the same for other `refresh_when_reload_*`.

empress_gui.py
Qing-Yang-1211: Ok!

empress_gui.py
Qing-Yang-1211: Done.

empress_gui.py
Qing-Yang-1211: Done.

empress_gui.py
Qing-Yang-1211: Done.

PR #129, opened 2020-07-11
Add pyinstaller to Pipfile, add freezing specs

ssantichaivekin:
Add `pyinstaller` dependency to pipfile, can be installed via `pipenv install --dev`. Add specs and readme for pyinstaller. The executables created by this pull request is provided in dtl group slack Friday July 10, 9:30 PM.

Pyinstaller packages need codesign for osx: https://github.com/pyinstaller/pyinstaller/wiki/Recipe-OSX-Code-Signing.

Tested by uploaded online and re-downloading it to run. It still works after I did that. Prof Ran also help tested the executable after code sign, so I think this is working as expected.

Haven't tested on linux/windows.


.github/workflows/integration-tests-and-draw-tests.yml
```diff
@@ -26,7 +26,6 @@ jobs:
         pipenv install --dev
     - name: Lint with flake8
       run: |
-        pipenv install flake8
```
ssantichaivekin, line 29: flake8 already installed via `pipenv install --dev`

PR #151, opened 2020-07-17
(outdated)GUI: separate reconciliation solutions and add toggle buttons for "One per cluster" + bring new windows to the front

Qing-Yang-1211:
When we choose "View reconciliations" and then use the "One per cluster" option, the buttons for toggling event frequency and internal node names are added.  Also, the gui is changed to pop up a separate window for each solution rather than displaying several of them in one window. I have tested the new features using the examples in the empress folder. 
Screen Shot 2020-07-17 at 4 37 03 PM
Issue: Only one "One per cluster" window can be closed when it becomes irrelevant (loading in new files/ compute reconciliations again). Ideally, all irrelevant "One per cluster" windows should be closed automatically as what the gui previously did. 

P.S. A minor improvement: always bring newly created tkinter windows to the front-most position for the user to see.
Issue: The new tkinter windows can be brought to the front when clicked on, but the main gui window will always stay behind all other newly created tkinter windows.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -1,8 +1,3 @@
-[[source]]
-url = "https://pypi.python.org/simple"
-verify_ssl = true
-name = "pypi"
-
 [requires]
 python_version = '3.7'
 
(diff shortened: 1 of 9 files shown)


Qing-Yang-1211:
Close in favor of #161.

PR #161, opened 2020-07-20
GUI: separate reconciliation solutions and add toggle buttons for "One per cluster" & bring new windows to the front

Qing-Yang-1211:
When we choose "View reconciliations" and then use the "One per cluster" option, the buttons for toggling event frequency and internal node names are added.  Also, the gui is changed to pop up a separate window for each solution rather than displaying several of them in one window. I have tested the new features using the examples in the empress folder. 
Screen Shot 2020-07-17 at 4 37 03 PM
Issue: Only one "One per cluster" window can be closed when it becomes irrelevant (loading in new files/ compute reconciliations again). Ideally, all irrelevant "One per cluster" windows should be closed automatically as what the gui previously did. 

P.S. A minor improvement: always bring newly created tkinter windows to the front-most position for the user to see.
Issue: The new tkinter windows can be brought to the front when clicked on, but the main gui window will always stay behind all other newly created tkinter windows.

### DIFF
--- a/empress_gui.py
+++ b/empress_gui.py
@@ -221,7 +221,7 @@ class App(tk.Frame):
         self.set_num_cluster_window = None
         self.view_solution_space_window_after_setting_clusters = None
         self.one_MPR_window = None
-        self.solution_window = None
+        self.solution_window_one_per_cluster = None
         self.view_pvalue_histogram_window = None
     
     def refresh_when_reload_host(self):
@@ -325,10 +325,8 @@ class App(tk.Frame):
         if self.one_MPR_window is not None and self.one_MPR_window.winfo_exists():
             self.one_MPR_window.destroy()
 
-        if self.num_cluster is not None:
-            for i in range(self.num_cluster):
-                if self.solution_window is not None and self.solution_window.winfo_exists():
-                    self.solution_window.destroy()
+        if self.solution_window_one_per_cluster is not None and self.solution_window_one_per_cluster.winfo_exists():
+            self.solution_window_one_per_cluster.destroy()
 
         if self.view_pvalue_histogram_window is not None and self.view_pvalue_histogram_window.winfo_exists():
             self.view_pvalue_histogram_window.destroy()
@@ -434,7 +432,7 @@ class App(tk.Frame):
         # Bring the new tkinter window to the front
         self.tanglegram_window.attributes('-topmost', True)
         self.tanglegram_window.focus_force()
-        self.tanglegram_window.bind('<FocusIn>', self.OnFocusIn)
+        self.tanglegram_window.bind('<FocusIn>', self.on_focus_in)
         # Creates a new frame
         tanglegram_frame = tk.Frame(self.tanglegram_window)
         tanglegram_frame.pack(fill=tk.BOTH, expand=1)
@@ -458,7 +456,7 @@ class App(tk.Frame):
         # Bring the new tkinter window to the front
         self.cost_space_window.attributes('-topmost', True)
         self.cost_space_window.focus_force()
-        self.cost_space_window.bind('<FocusIn>', self.OnFocusIn)
+        self.cost_space_window.bind('<FocusIn>', self.on_focus_in)
         # Creates a new frame
         plt_frame = tk.Frame(self.cost_space_window)
         plt_frame.pack(fill=tk.BOTH, expand=1)
@@ -651,7 +649,7 @@ class App(tk.Frame):
             # Bring the new tkinter window to the front
             self.entire_space_window.attributes('-topmost', True)
             self.entire_space_window.focus_force()
-            self.entire_space_window.bind('<FocusIn>', self.OnFocusIn)
+            self.entire_space_window.bind('<FocusIn>', self.on_focus_in)
             # Creates a new frame
             plt_frame = tk.Frame(self.entire_space_window)
             plt_frame.pack(fill=tk.BOTH, expand=1)
@@ -678,7 +676,7 @@ class App(tk.Frame):
         # Bring the new tkinter window to the front
         self.set_num_cluster_window.attributes('-topmost', True)
         self.set_num_cluster_window.focus_force()
-        self.set_num_cluster_window.bind('<FocusIn>', self.OnFocusIn)
+        self.set_num_cluster_window.bind('<FocusIn>', self.on_focus_in)
         # Creates a new frame
         self.set_num_cluster_frame = tk.Frame(self.set_num_cluster_window)
         self.set_num_cluster_frame.pack(fill=tk.BOTH, expand=tk.YES)
@@ -751,10 +749,8 @@ class App(tk.Frame):
         if self.view_solution_space_window_after_setting_clusters is not None and self.view_solution_space_window_after_setting_clusters.winfo_exists():
             self.view_solution_space_window_after_setting_clusters.destroy()
 
-        if self.num_cluster is not None:
-            for i in range(self.num_cluster):
-                if self.solution_window is not None and self.solution_window.winfo_exists():
-                    self.solution_window.destroy()
+        if self.solution_window_one_per_cluster is not None and self.solution_window_one_per_cluster.winfo_exists():
+            self.solution_window_one_per_cluster.destroy()
 
         if self.view_pvalue_histogram_window is not None and self.view_pvalue_histogram_window.winfo_exists():
             self.view_pvalue_histogram_window.destroy()
@@ -768,7 +764,7 @@ class App(tk.Frame):
             # Bring the new tkinter window to the front
             self.view_solution_space_window_after_setting_clusters.attributes('-topmost', True)
             self.view_solution_space_window_after_setting_clusters.focus_force()
-            self.view_solution_space_window_after_setting_clusters.bind('<FocusIn>', self.OnFocusIn)
+            self.view_solution_space_window_after_setting_clusters.bind('<FocusIn>', self.on_focus_in)
             SolutionSpaceWindow(self.view_solution_space_window_after_setting_clusters)
 
     def select_from_view_reconciliations_dropdown(self, event):
@@ -782,7 +778,7 @@ class App(tk.Frame):
             # Bring the new tkinter window to the front
             self.one_MPR_window.attributes('-topmost', True)
             self.one_MPR_window.focus_force()
-            self.one_MPR_window.bind('<FocusIn>', self.OnFocusIn)
+            self.one_MPR_window.bind('<FocusIn>', self.on_focus_in)
             ReconciliationsOneMPRWindow(self.one_MPR_window)
 
         elif self.view_reconciliations_var.get() == "One per cluster":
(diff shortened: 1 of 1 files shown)


empress_gui.py
```diff
@@ -219,9 +219,9 @@ def init_windows(self):
         self.cost_space_window = None
         self.entire_space_window = None
         self.set_num_cluster_window = None
+        self.view_solution_space_window_after_setting_clusters = None
```
ssantichaivekin, line 222: Why did you rename `self.view_solution_space_window` to `self.view_solution_space_window_after_setting_clusters`? Is it much different from `self.solution_window`?

empress_gui.py
```diff
@@ -319,14 +319,16 @@ def close_unnecessary_windows_if_opened(self):
...
+        if self.view_solution_space_window_after_setting_clusters is not None and self.view_solution_space_window_after_setting_clusters.winfo_exists():
+            self.view_solution_space_window_after_setting_clusters.destroy()
 
         if self.one_MPR_window is not None and self.one_MPR_window.winfo_exists():
             self.one_MPR_window.destroy()
 
-        if self.view_reconciliations_window is not None and self.view_reconciliations_window.winfo_exists():
-            self.view_reconciliations_window.destroy()
+        if self.num_cluster is not None:
+            for i in range(self.num_cluster):
```
ssantichaivekin, line 329: It seems like you are iterating from `i = 0` to `i = self.num_cluster - 1` but you never used the value `i`.

empress_gui.py
```diff
@@ -429,6 +431,10 @@ def display_tanglegram(self):
         self.tanglegram_window = tk.Toplevel(self.master)
         self.tanglegram_window.geometry("600x600")
         self.tanglegram_window.title("Tanglegram")
+        # Bring the new tkinter window to the front
+        self.tanglegram_window.attributes('-topmost', True)
```
ssantichaivekin, line 435: Do you need both `self.tanglegram_window.attributes('-topmost', True)` and `self.tanglegram_window.focus_force()`? Would only one of them work?

Also, if you copied this solution from online, please provide the link. (I can't tell whether you did).

empress_gui.py
```diff
@@ -429,6 +431,10 @@ def display_tanglegram(self):
...
+        self.tanglegram_window.focus_force()
+        self.tanglegram_window.bind('<FocusIn>', self.OnFocusIn)
```
ssantichaivekin, line 437: What is this line doing?

empress_gui.py
```diff
@@ -730,13 +748,29 @@ def compute_recon_solutions(self):
...
-        if self.view_solution_space_window is not None and self.view_solution_space_window.winfo_exists():
-            self.view_solution_space_window.destroy()
-        if self.view_reconciliations_window is not None and self.view_reconciliations_window.winfo_exists():
-            self.view_reconciliations_window.destroy()
+        if self.view_solution_space_window_after_setting_clusters is not None and self.view_solution_space_window_after_setting_clusters.winfo_exists():
+            self.view_solution_space_window_after_setting_clusters.destroy()
+
+        if self.num_cluster is not None:
+            for i in range(self.num_cluster):
+                if self.solution_window is not None and self.solution_window.winfo_exists():
```
ssantichaivekin, line 756: It seems like you are iterating from `i = 0` to `i = self.num_cluster - 1` but you never used the value `i`.

empress_gui.py
ssantichaivekin: Do you need a different solution window after setting clusters?

empress_gui.py
```diff
@@ -745,36 +779,48 @@ def select_from_view_reconciliations_dropdown(self, event):
...
     def open_window_reconciliations(self):
-        """Pop up a new tkinter window to display the reconciliations."""
+        """Pop up new tkinter windows to display one reconciliation per cluster."""
         if self.num_cluster is not None:
-            self.view_reconciliations_window = tk.Toplevel(self.master)
-            self.view_reconciliations_window.geometry("900x900")
-            self.view_reconciliations_window.title("View reconciliations")
-            ReconciliationsOnePerClusterWindow(self.view_reconciliations_window)
+            solution_number = 1
+            for solution in App.medians:
```
ssantichaivekin, line 796: Consider removing `solution_number = 1` and `solution_number = solution_number + 1` and write:
```python
for solution_index, solution in enumerate(App.medians):
    ...
    self.solution_window.title("View reconciliations " + str(solution_index + 1))
    ...
```

empress_gui.py
Qing-Yang-1211: self.view_solution_space_window_after_setting_clusters gives the solution space and self.solution_window gives the solutions for each cluster so they are different. I renamed self.view_solution_space_window to self.view_solution_space_window_after_setting_clusters because I got confused by the name without specifying that it is "after_setting_clusters" several times and it was kind of annoying.

Qing-Yang-1211: I think I can change self.solution_window to self.solution_window_one_per_cluster.

empress_gui.py
Qing-Yang-1211: Yes, because I wanted to create self.num_cluster number of new windows.

empress_gui.py
Qing-Yang-1211: I copied them from kakyo's answer in this post: https://stackoverflow.com/questions/1892339/how-to-make-a-tkinter-window-jump-to-the-front. self.tanglegram_window.attributes('-topmost', True) brings the new window to the front and self.tanglegram_window.focus_force() gives focus to that new window.

empress_gui.py
Qing-Yang-1211: When the user interacts with this window (i.e. bringing focus to it), this window will not be fixed in the front any more, so this window will only be brought to the front of all windows tentatively when it is opened instead of permanently staying in the front of all windows. This is achieved by calling the self.OnFocusIn function.

empress_gui.py
Qing-Yang-1211: Yes, because I wanted to delete the self.num_cluster number of tkinter windows created by "One per cluster".

empress_gui.py
Qing-Yang-1211: Yes, because I wanted to delete the self.num_cluster number of tkinter windows created by "One per cluster".

empress_gui.py
Qing-Yang-1211: Ok!

empress_gui.py
ssantichaivekin: I'm confused about what this code is doing. It seems to me that `self.solution_window` is one variable and you are attempting to destroy the same `self.solution_window` multiple number of times.

Qing-Yang-1211: Actually, this is not working currently and I will remove this. This issue is mentioned in the description of this pull request.

empress_gui.py
ssantichaivekin: Please add the link to the answer you copied from https://stackoverflow.com/a/53644859/2860949 as a comment in your code. This way, future maintainers of the codebase can understand you intentions and motivations behind this code.
Screen Shot 2020-07-20 at 11 19 21 AM

empress_gui.py
ssantichaivekin: Rename `self.OnFocusIn` to `self.bring_to_front`.

empress_gui.py
ssantichaivekin: Seems like you are destroying the same `self.solution_window` multiple times.

empress_gui.py
Qing-Yang-1211: I have changed the code. So currently only one window created by "One per cluster" can be closed when it becomes irrelevant (e.g. loading in new input files) regardless of how many new tkinter windows created by "One per cluster". I can create an issue for this if it is necessary.

empress_gui.py
Qing-Yang-1211: self.tanglegram_window.focus_force() is not necessary if we don't want to bring the focus to the newly created window automatically.

empress_gui.py
ssantichaivekin: Is there a way for the two variables to share the same code?

Also, I think it's generally good to have the variable name be more about what the variable is and less about when we need to use the variable. I'm still not sure what the code is doing, so I cannot suggest a new name.

Qing-Yang-1211: They cannot share the same code. You can see what they do in the two classes: class SolutionSpaceWindow (for self.view_solution_space_window_after_setting_clusters) and class ReconciliationsOnePerClusterWindow (for self.solution_window_one_per_cluster). SolutionSpaceWindow draws the solution space histogram and ReconciliationsOnePerClusterWindow draws the reconciliation graphs for clusters. 
This is an example of the solution space histogram:
Screen Shot 2020-07-21 at 2 37 55 AM

empress_gui.py
Qing-Yang-1211: Done! :))

empress_gui.py
Qing-Yang-1211: Ok!

empress_gui.py
Qing-Yang-1211: I have removed it.

Qing-Yang-1211:
> I cannot understand some of the code written here. Still, you are the team's expert on tkinter, so if you think it works I'm okay with it.
> 
> Also, I'm not sure I understand what you mean when you say
> 
> > Issue: Only one "One per cluster" window can be closed when it becomes irrelevant (loading in new files/ compute reconciliations again). Ideally, all irrelevant "One per cluster" windows should be closed automatically as what the gui previously did.
> 
> Could you elaborate on it more?

If the user loads in three input files and enters 3 to be the number of clusters, 3 new tkinter windows will pop up when the user click on "One per cluster". Ideally, if the user loads in three different input files now, all the 3 old tkinter windows created by "One per cluster" should be closed automatically. However, only one of them can be closed automatically now and it is always one because I don't have a way to refer to the three different tkinter windows created by "One per cluster" (and thus close all of them automatically).

empress_gui.py
ssantichaivekin: Do you think it would be good to name `self.view_solution_space_window_after_setting_clusters` to `self.solution_space_for_clusters_window` and `self.solution_window_one_per_cluster` to `self.view_reconciliations_window` or something of this sort?

empress_gui.py
ssantichaivekin: I'm not sure what the problem is, could you create a new issue for this with a number of screenshots?

empress_gui.py
Qing-Yang-1211: How about self.solution_space_for_clusters_window and self.view_reconciliations_for_clusters_window (since there is also "One MPR" under "Veiw reconciliations")? :))

ssantichaivekin:
I don't think the users will expect the pop-up windows to be closed when she uploads a new host/parasite/mapping. For this pull request, let's not close any of the "view reconciliation for each cluster" windows.

Qing-Yang-1211:
> I don't think users expect the pop-up windows to be closed when she uploads a new host/parasite/mapping. For this pull request, let's not close any of the "view reconciliation for each cluster" windows. Then, open an issue on this topic. We don't really have to solve this in the summer, but let's make sure we document it really well.

Please refer to issue #163 about this.

PR #170, opened 2020-07-22
Increase font size in reconciliation visualization

justinjiang1212:
Change font size so that it looks like this now:
Screen Shot 2020-07-21 at 4 39 47 PM

### DIFF
--- a/empress/recon_vis/recon_viewer.py
+++ b/empress/recon_vis/recon_viewer.py
@@ -378,7 +378,7 @@ def _sigmoid(x: float):
     :param x: A number to be plugged into the function
     :return a sigmoid value based on the input value, x
     """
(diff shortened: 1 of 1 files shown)


empress/recon_vis/recon_viewer.py
```diff
@@ -378,7 +378,7 @@ def _sigmoid(x: float):
     :param x: A number to be plugged into the function
     :return a sigmoid value based on the input value, x
     """
-    return (1 / (1 + math.e**(-x)))
+    return (1.5 / (1 + math.e**(-x)))
```
ssantichaivekin, line 381: It might be cleaner to write `1.5 * _sigmoid(x)` on line 372 and keep line 381 the same.

justinjiang1212:
Fixed

PR #176, opened 2020-07-23
remove the display optional argument from CLI cost-regions

JingyiRose:
We remove ``--display`` from the optional arguments of the ``cost-regions`` command of the command-line interface. The command-line interface should not display a pop-up menu. The user can use the GUI if he/she wants a pop-up.

The reason we set the matplotlib backend was that we want the pop-up to work correctly on the command-line interface. Now that we don't display a pop-up anymore, we can remove that line.

### DIFF
--- a/empress/__init__.py
+++ b/empress/__init__.py
@@ -2,7 +2,6 @@
 Wraps empress functionalities
 """
 from typing import Dict
-import matplotlib
 import sys
 
 from matplotlib import pyplot as plt
--- a/empress/__init__.py
+++ b/empress/__init__.py
@@ -4,13 +4,7 @@ Wraps empress functionalities
 from typing import Dict
 import matplotlib
 import sys
-# the tkagg backend is for pop-up windows, and will not work in environments
-# without graphics such as a remote server. Refer to issue #49
-try:
-    matplotlib.use("tkagg")
-except ImportError:
-    print("Using Agg backend: will not be able to create pop-up windows.", file=sys.stderr)
-    matplotlib.use("Agg")
+
 from matplotlib import pyplot as plt
 from typing import List, Tuple
 from abc import ABC, abstractmethod
--- a/empress/xscape/costscape.py
+++ b/empress/xscape/costscape.py
@@ -27,6 +27,6 @@ def solve(newick_data, transferMin, transferMax, dupMin, dupMax, optional):
     print("Elapsed time %.2f seconds" % elapsedTime)
     plotcosts.plotcosts(CVlist, transferMin, transferMax, dupMin, dupMax, \
                         optional.outfile, \
-                        optional.log, False)
+                        optional.log)
     if optional.outfile != "":
         print("Output written to file: ", optional.outfile)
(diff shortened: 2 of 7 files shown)


empress/xscape/costscape.py
```diff
@@ -31,6 +27,6 @@ def solve(newick_data, transferMin, transferMax, dupMin, dupMax, optional):
     print("Elapsed time %.2f seconds" % elapsedTime)
     plotcosts.plotcosts(CVlist, transferMin, transferMax, dupMin, dupMax, \
                         optional.outfile, \
-                        optional.log, display)
+                        optional.log, False)
```
ssantichaivekin, line 30: Let's go into `plotcosts.plotcosts` and remove the display argument from the code.

empress_cli.py
```diff
@@ -6,6 +6,10 @@
 
 import argparse
 
+import matplotlib
+# use agg backend, see # https://matplotlib.org/faq/usage_faq.html#what-is-a-backend
+matplotlib.use("agg")
```
ssantichaivekin, line 11: Let's open an issue that talks about how tkinter backend breaks your computer. Then, write another line of comment which says "agg" is a non-interactive backend which cannot pop up a window
```python
# "agg" is a non-interactive backend which cannot pop up a window
# we don't need to pop up a window in the clip, there is a tkinter bug which occurs when we 
# pop up a matplotlib window from the cli.
# see [link to the issue]
```

empress_gui.py
```diff
@@ -9,7 +9,10 @@
 import sys
 import pathlib
 
-import matplotlib.pyplot as plt
+import matplotlib
+# use tkinter backend, see # https://matplotlib.org/faq/usage_faq.html#what-is-a-backend
```
ssantichaivekin, line 13: Let's write
```
# use tkinter backend, required for embedding in Tk canvas
# see https://matplotlib.org/faq/usage_faq.html#what-is-a-backend
```

empress_cli.py
ssantichaivekin: Actually let's just remove the whole thing and document it in the pull request description.

empress_gui.py
```diff
@@ -10,8 +10,8 @@
 import pathlib
 
 import matplotlib
-# use tkinter backend, see # https://matplotlib.org/faq/usage_faq.html#what-is-a-backend
-matplotlib.use("tkagg")
+# use tkinter backend, required for embedding in Tk canvas
```
ssantichaivekin, line 13: Please remove these comments too. We are not using tkinter backend for matplotlib anymore.

empress_gui.py
```diff
@@ -9,7 +9,10 @@
 import sys
 import pathlib
 
-import matplotlib.pyplot as plt
+import matplotlib
```
ssantichaivekin, line 12: And remove `import matplotlib` line to as it is already import in `from matplotlib import pyplot as plt`

empress/__init__.py
```diff
@@ -4,13 +4,7 @@
 from typing import Dict
 import matplotlib
```
ssantichaivekin, line 5: Remove `import matplotlib` line.

PR #177, opened 2020-07-23
GUI: fix the bug about the cluster window & the p-value histogram button 

### LINKED ISSUE #180: GUI: Clusters and P-value Histogram buttons not working on Vidua-Finches data
**Problem:**

Have to click on `View solution space` -> `Clusters` several times before the `Set the number of clusters` window would open. P-Value histogram not showing up.

**Steps to reproduce:**

1. Load the host/parasite/mapping of the gopher-louse data to empress GUI.
2. Change transfer cost from 1 to 1.5
3. Click `Compute Reconciliations`. The lower three buttons become available.
4. Click `View solution space` -> `Clusters`. **Nothing pops up**. `View reconciliations` and `View p-value histogram` **suddenly becomes grayed out**. The expected behavior is that the `Set the number of clusters` window pops up and two buttons remain available.
5. Click on `View solution space` -> `Clusters` again. `View reconciliations` becomes available but `View p-value histogram` remains unavailable.

This bug is discovered by Prof Ran.

Qing-Yang-1211:
Resolves #180 
This bug occurs to all data, including small data too. 
This pull request addresses the bug discovered by Prof Ran during testing:
"When I want to cluster data, sometimes I have to click on the cluster option a few times before the "Set the number of clusters" windows opens. In particular, I tried this on the vidua data. I set duplication cost to 1.0, transfer to 1.5, and loss to 1.0. Then, I clicked "Compute reconcilations". Then, I went to "View solution space" and had to click on "Clusters" several times before the "Number of clusters" window would open. Moreover, the p-value histogram button is grayed out and I can't activate it in this case. Vidua is the the Finches data."

Both issues were addressed and tested using the Finches data. The cluster window will pop up on the first time of clicking on "Number of clusters" and the p-value histogram button will not be grayed out in this case. However, another change is that when the user changes dtl costs in the Entry widgets, the self.view_solution_space_dropdown, self.view_reconciliations_dropdown, and self.view_pvalue_histogram_btn will no longer be disabled automatically. They will remain accessible and the results associated with them will be updated once the user clicks on "Compute reconciliations".

### DIFF
--- a/empress_gui.py
+++ b/empress_gui.py
@@ -582,13 +582,17 @@ class App(tk.Frame):
         if self.dup_cost is not None and self.trans_cost is not None and self.loss_cost is not None:
             # Enable the next button
             self.compute_reconciliations_btn.configure(state=tk.NORMAL)
-            self.disable_unaccessable_widgets()
             self.close_unnecessary_windows_if_opened()
         else:
             self.compute_reconciliations_btn.configure(state=tk.DISABLED)
 
     def display_recon_information(self):
         """Display numeric reconciliation results and close unnecessary windows."""
+        self.master.focus()  # remove focus from Entry and thus stop its validate command
+        self.close_unnecessary_windows_if_opened()
+        self.view_solution_space_dropdown.configure(state=tk.NORMAL)
+        self.view_reconciliations_dropdown.configure(state=tk.NORMAL)
+        self.view_pvalue_histogram_btn.configure(state=tk.NORMAL)
         App.recon_graph = self.recon_input.reconcile(self.dup_cost, self.trans_cost, self.loss_cost)
         self.recon_count = App.recon_graph.n_recon
         self.cospec_count, self.dup_count, self.trans_count, self.loss_count = App.recon_graph.median().count_events()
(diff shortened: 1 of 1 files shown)


Qing-Yang-1211:
> That was a fast fix! 😀 It seems that the fix is very small too, which is very nice! However, I still don't understand the bug occurs and how this change fixes the bug. Could you explain a little further why the bug occurs and how this change fixes the bug? Also, don't forget to write "Resolves #issue-number" at the end of the pull request!

I figured out these issues using the debugger. "self.master.focus()  # remove focus from Entry and thus stop its validate command" removes the focus from the last Entry widget edited when the user clicks on "Compute reconciliations" so that the validate command associated with the Entry widget is stopped tentatively and the update_compute_reconciliations_btn function won't be called constantly. This will help to pop up the cluster window when the user clicks on "Number of clusters" for the first time, instead of twice (due to the extra calling of the update_compute_reconciliations_btn function). 
The removal of "self.disable_unaccessable_widgets()" in the update_compute_reconciliations_btn function solves the problem that the p-value button is not enabled in the aforementioned test case.

ssantichaivekin:
I can understand why `self.master.focus()` stops the validate command of child windows from being run. Still, if we are not modifying the text, the validate command of the child window shouldn't run anyway. Is there something unexpected going on here?

ssantichaivekin:
But when I try the changes, it does fix the problem. Great job Qing! 😀

Qing-Yang-1211:
> I can understand why `self.master.focus()` stops the validate command of child windows from being run. Still, if we are not modifying the text, the validate command of the child window shouldn't run anyway. Is there something unexpected going on here?

I think it should run even if we are not editing the text because we are validating on "All" changes, which includes "FocusIn". Thus, removing focus could solve the problem.

ssantichaivekin:
From the Files Changed, it seems that you are not removing the focus.

ssantichaivekin:
Closing in favor of #218

PR #183, opened 2020-07-24
GUI: fix the updating num_cluster bug

### LINKED ISSUE #182: GUI: Number of clusters and old cluster data not removed when changing cost
**Problem:**

The number of clusters is not updated when using `Set the number of clusters` after changing the cost of duplication/transfer/loss. The previous clustering data also remains available after changing the cost. Clicking on it displays the old clustering data.

Screen Shot 2020-07-23 at 8 52 02 PM

**Steps to reproduce:**

1. Load gopher-louse data from examples.

2. Click `Compute Reconciliations` without changing the costs (it's 1,1,1). The number of MPRs is two.

3. Click `View solution space` -> `Clusters`. Enter 1.
Screen Shot 2020-07-23 at 9 55 17 PM

4. Click `View reconciliations` -> `One per cluster`. It shows one reconciliation drawing. Nothing is wrong so far.
Screen Shot 2020-07-23 at 9 55 41 PM

5. Change duplication cost to 2 and transfer cost to 4, then click compute reconciliations. There are 2 MPRs.

6. Click `View solution space` -> `Clusters`. Enter 2. It produces incorrect output.
Screen Shot 2020-07-23 at 8 52 02 PM

7. Click `View reconciliations` -> `One per cluster` shows the old reconciliation.
Screen Shot 2020-07-23 at 10 05 29 PM

Qing-Yang-1211:
Resolves #182.
The problem was that when creating the self.set_num_cluster_window, I shouldn't have validated self.num_cluster_input=1 by default every time, and should instead use self.num_cluster_entry_box.get() to get the current value in the entry field and validate that number. This also works if the user don't change the default value in the entry field since self.num_cluster_input was initialized to be 1 and the number in the entry field is 1 to start with.

### DIFF
--- a/empress_gui.py
+++ b/empress_gui.py
@@ -694,7 +694,7 @@ class App(tk.Frame):
         self.num_cluster_entry_box = CustomEntry(self.set_num_cluster_frame, width=3, textvariable=self.num_cluster_input)
         self.num_cluster_entry_box.set_border_color("green")
         self.num_cluster_entry_box.validate(validate="key", validatecommand=num_cluster_vcmd)
-        self.validate_num_cluster_input(1)
+        self.validate_num_cluster_input(self.num_cluster_entry_box.get())
         self.num_cluster_label.grid(row=0, column=0, sticky="e")
         self.num_cluster_entry_box.grid(row=0, column=1, sticky="w")
         self.num_cluster_error.grid(row=0, column=2)


ssantichaivekin:
This only resolve part of the problem in #182.

ssantichaivekin:
Closing in favor of #219 .

PR #193, opened 2020-07-31
Change legend display location of cost-regions and reconciliation

rlhadas:
in costscape (upper right) and reconcilation  visualization (upper left).


ssantichaivekin:
Closing in favor of #190

ssantichaivekin:
Reopen because #190 is actually unrelated

PR #199, opened 2020-08-03
GUI: Change clustering to use n-splits

aremath:
Previously, clustering used depth with a default of 4. Now it uses
n-splits with a default of 16, set by a variable. The primary result of
this change is that the user will never get fewer clusters than
requested. It may either improve or degrade the clustering quality in
different cases.


rlhadas:
Looks good!

PR #207, opened 2020-08-04
Improves histogram cli

aremath:
Makes the .csv argument mandatory, merges some shared code with reconcile.

### DIFF
--- a/cli_commands/_shared_utils.py
+++ b/cli_commands/_shared_utils.py
@@ -1,4 +1,5 @@
 import argparse
+from pathlib import Path
 
 def add_recon_input_args_to_parser(parser: argparse.ArgumentParser):
     parser.add_argument("host", metavar="<host_file>",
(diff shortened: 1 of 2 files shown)


ssantichaivekin:
It seems like the changes in this pull request is included in #206 . I'm closing this pull request in favor of #206

aremath:
Not all of the changes are included in #206. They both share a new function in `cli_commands/_shared_utils.py`, but this branch also makes other changes to `cli_commands/histogram.py`

ssantichaivekin:
Okay. I will review this pull request after the other one is merged into master then. If you want me to review both pull requests at the same time, please remove duplicate changes from the two pull requests. If the duplicate changes are necessary, instead of creating both pull requests to master, create one pull request `master` <- `branch-a`, and then another pull request `branch-a` <- `branch-b` so that changes are not displayed twice.

ssantichaivekin:
Could you explain why we are making the csv argument mandatory? Also, if `nargs` is not specified, the argument is optional by default. It seems like the pull request is not making the argument mandatory.

PR #209, opened 2020-08-05
Add --n-samples cli option to p-value command

aremath:
Adds a way to set the number of samples used to generate the p-value histogram in the cli.


ssantichaivekin:
I will modify and merge this.

PR #210, opened 2020-08-05
Minor histogram readme update

aremath:
Updates Readme to document the histogram .csv format.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -176,6 +176,6 @@ $ python empress_cli.py p-value examples/heliconius_host.nwk examples/heliconius
 
 View a tanglegram of the given files, showing the tip mapping.
 ```bash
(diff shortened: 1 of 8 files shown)


Readme.md
```diff
@@ -140,6 +140,7 @@ For example, to run Pairwise Distance Histogram that outputs a csv file at `foo.
 $ python empress_cli.py histogram examples/heliconius_host.nwk examples/heliconius_parasite.nwk \
                                   examples/heliconius_mapping.mapping --csv foo.csv --histogram bar.pdf --ynorm
 ```
+CSV output has rows with two values each. The first value on each row is the distance (measured using the symmetric set difference between the event sets), and the second value is the number of MPR pairs that are that far apart from each other.
```
ssantichaivekin, line 143: Do you think it would be a good idea if we also provide an example with an explanation?

ssantichaivekin: Actually I don't think this is needed.

PR #219, opened 2020-08-08
[PART 2] GUI: Fix old clusters do not get replaced bug

### LINKED ISSUE #182: GUI: Number of clusters and old cluster data not removed when changing cost

ssantichaivekin:
Fixes two bugs in #182

1. Incorrect number of clusters calculated after changing the cost.

This is because the code assumes that the input box of the cluster window always has 1 as its starting value.  However, this is not always the case since if we change the cost and click on the cluster window again, the current behavior is that the box will have value as what the box has last time it is used.

```python
self.validate_num_cluster_input(1)  # old
self.validate_num_cluster_input(self.num_cluster_entry_box.get())  # new
```

2. `View reconciliations` -> `One per cluster` is not disabled properly, it can display old information.

Fixed by disabling the dropdown option when we change the cost and re-enabling them when we click on Enter button.

Resolve #182

### DIFF
--- a/empress_gui.py
+++ b/empress_gui.py
@@ -205,7 +205,7 @@ class App(tk.Frame):
             *self.view_reconciliations_options, command=self.select_from_view_reconciliations_dropdown)
         self.view_reconciliations_dropdown.configure(width=15)
         self.view_reconciliations_dropdown.configure(state=tk.DISABLED)
-        self.view_reconciliations_dropdown['menu'].entryconfigure("One per cluster", state = "disabled")
+        self.view_reconciliations_dropdown['menu'].entryconfigure("One per cluster", state="disabled")
         self.view_reconciliations_dropdown.grid(row=5, column=0)
 
     def init_view_pvalue_histogram(self):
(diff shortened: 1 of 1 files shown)


ssantichaivekin:
PS. You could review (and approve) this pull request even when the base branch is not master.

PR #224, opened 2020-08-09
GUI: Fix toolbar error when changing reconciliation boolean options

### LINKED ISSUE #223: GUI: empty reconciliation window when changing boolean options in cluster
Steps to reproduce:
1. Load gopher-louse data set with the costs of 1,1,1
2. Click `Compute reconciliations`
3. Click `View solution space`, leave the number of clusters to 1, and press `Enter` (note that the number of clusters does not matter. This happens no matter what number of clusters you choose).
4. Click `View reconciliations` -> `one per cluster`
5. Check and uncheck `Display internal node names`
6. The screen turns white and the following message appears on the console.
Screen Shot 2020-08-09 at 11 00 02 PM

```python
Exception in Tkinter callback
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/tkinter/__init__.py", line 1705, in __call__
    return self.func(*args)
  File "empress_gui.py", line 956, in update_median_recons
    self.toolbar.destroy()
  File "/Users/santisantichaivekin/.local/share/virtualenvs/empress-dCwpO06i/lib/python3.7/site-packages/matplotlib/backends/_backend_tk.py", line 531, in destroy
    del self.message
AttributeError: message
```
I believe this happens regardless of the costs. However, the same error doesn't happen if we display only one reconciliation and don't use clusters.

ssantichaivekin:
Problem: the matplotlib toolbar could not be deleted properly, so it raises an exception. This is because when we update the toolbar for the "one per reconciliation windows," it creates a new toolbar using the code
```python
toolbar = NavigationToolbar2Tk(self.canvas, self.frame)`
```
This toolbar variable disappears after the update function finishes. However, when we remove the toolbar, we call `self.toolbar.destroy()`.

Solution: tie the toolbar variable to self at creation
```python
self.toolbar = NavigationToolbar2Tk(self.canvas, self.frame)`
```

Tested on macOS Catalina 10.15.5, can check/uncheck boolean options in #223 correctly.

Resolves #223

### DIFF
--- a/empress_gui.py
+++ b/empress_gui.py
@@ -962,8 +962,8 @@ class ReconciliationsOnePerClusterWindow(tk.Frame):
         self.canvas.draw()
         self.canvas.get_tk_widget().pack(side=tk.TOP, fill=tk.BOTH, expand=True)
         # Recreate the toolbar
-        toolbar = NavigationToolbar2Tk(self.canvas, self.frame)
-        toolbar.update()
+        self.toolbar = NavigationToolbar2Tk(self.canvas, self.frame)
+        self.toolbar.update()
 
 # p-value Histogram
 class PValueHistogramWindow(tk.Frame):


ssantichaivekin:
If you are merging this, could you modify the commit message of the merge to be more explanatory than just "fix bugs"? Feel free to copy from the pull request description.

Issue #229, opened 2021-06-02
Cannot export reconciliation images as pdf


Jorge-Dona:
Hi,

I'm trying to save a figure in a vectorial format (e.g., PDF).  However, I'm only allowed to save figures in PNG.  I'm running eMPRess on a Mac and installed it following the instructions from the eMPRess website –other than this is running with no issues.

<img width="2075" alt="Screen Shot 2021-06-02 at 9 11 46 AM" src="https://user-images.githubusercontent.com/9450421/120495757-9bae7a80-c382-11eb-9ed5-db8044b942d7.png">

Thanks!

Jorge

ssantichaivekin:
Hi Jorge,

Thank you for using eMPRess!

It seems like my save dialogue box looks different from yours. This might be due to the macOS version. Here's what the save dialogue box looks like for me:

<img width="843" alt="Screen Shot 2021-06-02 at 10 06 23 PM" src="https://user-images.githubusercontent.com/19219105/120504990-c7ffe080-c3ee-11eb-975c-b096c643a700.png">

My eMPRess doesn't show the format of the file being saved. However, if I change the filename to `image.pdf` the image saves correctly. Could you try that and see how it goes?

Jorge-Dona:
Hi Santi.

I tried that already and it didn't work.  When I do that I get this:

![image](https://user-images.githubusercontent.com/9450421/120533703-a4b14300-c3a6-11eb-82e4-c1660e45d970.png)

Then, I've tried both the "use .png" option and "use both" and in both cases, the saved figure is a .png file. My macOS version is 11.2.3 Big Sur.

Thanks!

Jorge

ssantichaivekin:
I will try to work on some fixes, but I don't have a Big Sur machine so it might take a while. Meanwhile, you can send me the files at jsantichaivekin@g.hmc.edu and I will send back the pdf.

ssantichaivekin:
Could you try using [this version of eMPRess](https://github.com/ssantichaivekin/empress/suites/2898651901/artifacts/64949852) instead and see how it goes?

Jorge-Dona:
Hi Santi,

It worked.  I installed it using the command line.

Thanks!

Jorge

ssantichaivekin:
No problem! A little heads up: that version doesn't work with macOS dark mode. If you switch to dark mode you might see the texts disappearing.

ssantichaivekin:
[...]

PR #230, opened 2021-06-03
Update pipfile & pipfile.lock

ssantichaivekin:
Jorge reports in #229 that the executable on the website doesn't work with macOS Big Sur. I downloaded and confirmed with a Big Sur machine that it didn't work. However, updating dependencies fix this problem for unknown reasons. I tested the executable produced by 01e6a1f and it runs without a problem. It also runs correctly on Jorge's computer.

In addition to updating dependencies:
- Added `pipenv lock --keep-outdated` to automated workflows to support cross-platform deployment. See pypa/pipenv#3902.


rlhadas:
HI Santi.  I noticed that the Windows build failed.

ssantichaivekin:
@rlhadas I also noticed that and haven't merge this change 😢

PR #235, opened 2021-06-06
Allow macOS notarization

ssantichaivekin:
Add `com.apple.security.cs.allow-unsigned-executable-memory = True` to `info.plist` to allow macOS notarization.

 See https://haim.dev/posts/2020-08-08-python-macos-app/


ssantichaivekin:
Turns out this is not the right way to do it.

PR #243, opened 2021-06-12
GUI: Set default save format to .pdf

ssantichaivekin:
Empress GUI now defaults to saving images as `.pdf` instead of `.png`. The pdf format is more helpful because it's a vector format that can be zoomed in infinitely.

### DIFF
--- a/empress_gui.py
+++ b/empress_gui.py
@@ -9,11 +9,12 @@ import os
 import sys
 import pathlib
 
+import matplotlib
+matplotlib.rcParams['savefig.format'] = 'pdf'  # default save format
 from matplotlib import pyplot as plt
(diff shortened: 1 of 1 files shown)


empress_gui.py
```diff
@@ -9,11 +9,12 @@
 import sys
 import pathlib
 
+import matplotlib
+matplotlib.rcParams['savefig.format'] = 'pdf'  # default save format
 from matplotlib import pyplot as plt
 from matplotlib.backends.backend_tkagg import FigureCanvasTkAgg, NavigationToolbar2Tk
 
 import empress
-from empress import input_reader
```
ssantichaivekin, line 16: Remove unused import

PR #244, opened 2021-06-12
Add executable permission to linux app

ssantichaivekin:
Add executable permission to eMPRess app on Linux so that users don't have to run  right-click on **empress_gui** file, and click Properties, then select the Permissions tab. Then, check the box that says Allow executing file as a program. (or run `chmod +x ./empress_gui`)


ssantichaivekin:
Doesn't work.

